PR #2, opened 2015-09-25
Added options for PAL capture

privatezero:
Hi, hope I am using the GitHub interface correctly!
Since we have a bunch of PAL here at the UW, I worked a bit on this to add options for PAL capture. Thought I would send them along!


dericed:
LGTM, thanks.

jfarbowitz:
Have you tested this to verify that it works properly?

privatezero:
I verified everything for capture and it checked out ok- today I tried capture and passthrough actually using a physical PAL tape and I noticed an error when using passthrough mode- will fix shortly!  (Capture worked fine).

PR #45, opened 2016-02-23
use geq to highlight max/min val and out of range vals

dericed:


richardpl:
geq is slooow.

dericed:
lutyuv?

richardpl:
Yes, also geq is 8bit only.

retokromer:
Ping! This is approaching one year…

dericed:
dropping this, the idea could be redone in lutyuv but as noted geq does slow it down too much

retokromer:
@dericed: OK, thanks. I’ll check if there is any idea in de-restorer that may be recycled for improving speed here.

retokromer:
@dericed : This branch is still alive. Do you plan to replace the `geq` filter by `lutyuv`? Or should this be moved into a "wish" issue?

PR #49, opened 2016-05-18
Digibase integration

jfarbowitz:


vrecord
```diff
@@ -370,6 +371,18 @@ lookup_qctoolsxml(){
...
 }
 
+lookup_db_choice(){
+    case "${1}" in
+        "Yes")
+            ;;
+        "No")
+            ;;
+        Q) exit 0 ;;
+        q) exit 0 ;;
```
retokromer: Either `Q|q) exit 0 ;;` or `[Qq]) exit 0 ;;` would be one line shorter

retokromer:
What is the status of this?

retokromer:
Ping!

dericed:
I recommend closing for the integration with digibase and superseding digibase's use with something similar to @privatezero's db integration in ltopers and mm. However there are non-digibase changes mixed in here which should be reviewed.

retokromer:
The `audiomap` has also been changed in the master.

PR #58, opened 2016-08-10
Created check for recording time compliance

privatezero:
will warn and exit if recording time value is not a number

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -730,6 +730,16 @@ if [[ "${runtype}" = "edit" ]] ; then
     rm $pashua_configfile
     if [[ "$cb" == 0 ]] ; then
         # report back options
+        
+#confirm legal value for diration
+if [ -n "${duration}" ]; then
+real_regexp="[+-]?([0-9]+|[0-9]+\.[0-9]*|\.[0-9]+)"
+if [[ $duration =~ ^$real_regexp$ ]]; then
+is_number=yes
+else echo "Illegal value for recording time. Input must only be numbers." && exit
+fi
+fi
+        # report back options
         echo "Variables set:"
         echo "  dir = ${dir}"
         echo "  logdir = ${logdir}"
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -730,6 +730,16 @@ if [[ "${runtype}" = "edit" ]] ; then
     rm $pashua_configfile
     if [[ "$cb" == 0 ]] ; then
         # report back options
+        
+#confirm legal value for diration
```
retokromer: typo: duration

vrecord
```diff
@@ -925,6 +935,15 @@ if [ -z "${duration}" ] ; then
...
 
+# confirm legal value for diration
+if [ -n "${duration}" ]; then
+real_regexp="[+-]?([0-9]+|[0-9]+\.[0-9]*|\.[0-9]+)"
+if [[ $duration =~ ^$real_regexp$ ]]; then
+is_number=yes
+else echo "Illegal value for recording time. Input must only be numbers." && exit
+fi
+fi
+
```
retokromer: Why two time the same function?

privatezero: The repeat of the function accommodates the different ways a user could enter the duration value either through the GUI or the CLI prompt.  If they leave the option blank in the GUI then vrecord will prompt the user for a value via CLI (and verify its validity) after they have input the identifier for the file to be created.

Having the function also after the GUI outputs means that if an illegal value is entered via the GUI vrecord will immediately warn the user rather than proceeding to ask for a file identifier and then aborting.

retokromer: Thank you for the answer! I did understand that. My remark was meant from the maintainer's perspective: only once the code, called in two different ways.

privatezero: Haha! You are correct- I certainly can rework that to make it more efficient!

retokromer: Sorry, my programming style comes from the last millennium: the first computer I built, 41 years ago, had 256 B (no letter missing!) of memory...

PR #81, opened 2016-10-27
BRNG warning

savcampbell:
A warning will now appear (in cow form) if the file contains frames
with pixels that fall out of broadcast range


dericed:
LGTM though after running it a bit I suspect we may adjust the default BRNG_OUTLIER_THRSHLD. I expect that >14 frames with 3% out of broadcast range might occur on nearly every recording.

PR #90, opened 2016-11-16
Muxer options

### LINKED ISSUE #82: add muxer options
To make the recordings more self-description there's some options we should consider using.

For QuickTime add `-movflags write_colr` will write colorspace data, though am unsure if the current nut pipe into ffmpeg would probably convey the write colorspace data, it may need to be declared as well.

Similarly if mxf is used, we could consider adding `-signal_standard  bt601` if the input is known to be bt601 (such as the NTSC and PAL options).

For matroska, IIUC the color data will write if known, but it may need to be declared.

savcampbell:
Colorspace metadata now available for .mkv and .mov files. Fixes #82

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Vrecord Documentation #
-This documentation is up to date as of vrecord v0.7.10 (Released on 12/10/15)
+This documentation is up to date as of vrecord v0.7.10 (Released on 2015–12–10)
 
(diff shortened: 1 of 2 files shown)


dericed:
I added some comments, if you need help rebasing and squashing, lemme know.

dericed:
thx!

PR #92, opened 2016-11-17
Warning for incompatible codecs/containers 

### LINKED ISSUE #15: Incompatible codecs/containers selectable
From my experience, FFv1 (and possibly v210 and prores?) are not currently compatible with .MXF. 
Perhaps a warning could pop up if these options are selected when using the CLI. I'm not sure how difficult it would be to prevent a user from selecting MXF in the GUI if an incompatible codec is selected?

savcampbell:
Fixes #15


kieranjol:
Can't test this but thanks for adding it!

dericed:
LGTM though it will still fail once it gets to ffmpeg

PR #95, opened 2016-11-18
housekeeping

retokromer:
- "modern" shebang
- tried to uniform indent to 4 and spacing
- $(...) preferred over \`...\`
- = preferred over == in [[...]]

Note: I am not sure that the additional `. "${config_file}"` in line 871 is necessary. I guess not, and, if so, I suggest to cancel and simplify this too.


dericed:
LGTM, did you test this?

retokromer:
As long as I can, yes (VHS only, and an old Blackmagic...).

PR #97, opened 2016-11-18
Handling of parameters passed at CLI (& some housekeeping en passant)

retokromer:
- deleted double `. "${config_file}"`
- and corrected some missed alignments

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+
+version="2016-11-19"
 config_file="$HOME/.$(basename "${0}").conf"
 
 inputoptions=""
@@ -177,7 +179,7 @@ report(){
     local echoopt=""
     local log_message=""
     OPTIND=1
-    while getopts ":qdwstn" opt ; do
+    while getopts "qdwstn" opt ; do
         case "${opt}" in
             q) color="${BLUE}" ;;                         # question mode, use color blue
             d) color="${GREEN}" ;;                        # declaration mode, use color green
--- a/vrecord
+++ b/vrecord
@@ -879,7 +879,7 @@ duration_check
         report -d "Bye then"
         exit
     elif [[ "$aftereditresponse" = "p" ]] ; then
-	runtype="passthrough"
+        runtype="passthrough"
     fi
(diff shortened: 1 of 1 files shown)


retokromer:
https://github.com/amiaopensource/vrecord/issues/98 still occurs. I am delving into this.

retokromer:
It’s looping at https://github.com/amiaopensource/vrecord/blob/master/vrecord#L227

vrecord
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+
+version="2016-11-19"
```
dericed: what is the recommended maintenance with this?

dericed:
I do wonder how to deal with vrecord versioning, adding `version="2016-11-19"` seems imply that contributors should change the `version` with any change. I'd rather move work on versioning handling into a separate PR as otherwise the PR looks good to me.

vrecord
retokromer: No idea! A date is better than an empty variable…

But we can also delete the printing of this variable at https://github.com/amiaopensource/vrecord/blob/b04905def37a5476a144b139981bba96bf23e57a/vrecord#L24

dericed: preferred

retokromer:
Yes, it was meant to be changed manually by every single contributor, therefore I put it at the beginning (a comment after would be useful… my fault), but I will delete it for this PR. And we can see this later.

retokromer:
Thank you @dericed and @privatezero !

privatezero:
Hi! I don't know if it would break interoperability or cause other problems, but in the branch I was working on for adding a secondary GUI I was adding in versioning information that pulled the version number from the install path used by Homebrew.  Basically it would take the directory name that the current version was installed to and use that as version.  Not sure if this is an efficient solution?

https://github.com/amiaopensource/vrecord/blob/add-secondary-gui/vrecord#L3

retokromer:
I’m always* in favour of versioning, and I strongly support your solution!

*see https://twitter.com/retokromer/status/777559276722196482 (today I would say RFC 3339).

PR #100, opened 2016-11-19
remove shotcutsavesas

dericed:
Not really in scope of the repo.


retokromer:
Yesterday I was indeed wondering what this was meant to achieve…!

Issue #96, opened 2016-11-18
issue with warning on saturation levels when 10 bit


dericed:
As reported by @savcampbell, since the signalstats filter is updated to 10 bit, the test at https://github.com/amiaopensource/vrecord/blob/5c462a99fdaa814983ca5a51b99124790051756e/vrecord#L1148 and https://github.com/amiaopensource/vrecord/blob/5c462a99fdaa814983ca5a51b99124790051756e/vrecord#L1201-L1203 is out of context in 10 bit. The `124` threshold in there should be contextually adjusted for 10 bit to `124 * 4` while 8 bit would stay at `124`.

dericed:
closed in https://github.com/amiaopensource/vrecord/pull/103

PR #103, opened 2016-11-22
10 bit saturation warning

savcampbell:
Saturation warning adjusted to report on illegal values in both 8-bit and 10-bit captures.


savcampbell:
Fixes #96

dericed:
LGTM

PR #105, opened 2016-11-23
Metadata for full range mkv

### LINKED ISSUE #87: Add metadata for full range recordings
Since vrecord now has a playback view option for full range capture, this should be reflected in metadata of the resulting video files. Metadata should specify whether a video file should be interpreted as broadcast range or full range.

savcampbell:
mkv captures can now be interpreted as Broadcast Range or Full Range
depending on selected playback view. Fixes #87

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1105,9 +1105,7 @@ else
     recordingfilter="[0:v:0]setfield=bff,setsar=40/27,setdar=4/3[vid1]"
 fi
 
-if [[ "${playbackview_choice}" = "Broadcast Range Visual" ]] ; then
-    middleoptions+=(-color_range mpeg)
-elif [[ "${playbackview_choice}" = "Full Range Visual" ]] ; then
+if [[ "${playbackview_choice}" = "Full Range Visual" ]] ; then
     middleoptions+=(-color_range jpeg)
 else
     middleoptions+=(-color_range mpeg)
--- a/vrecord
+++ b/vrecord
@@ -1096,15 +1096,21 @@ if [[ "${standard_choice}" = "PAL" ]] ; then
     middleoptions+=(-color_primaries bt470bg)
     middleoptions+=(-color_trc bt709)
     middleoptions+=(-colorspace bt470bg)
-    middleoptions+=(-color_range mpeg)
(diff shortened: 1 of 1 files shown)


dericed:
I wonder if these should be two separate options. @kellyhaydon do you want to record a full range file while you use a full range waveform viewer or just record a broadcast range file while you use a full range waveform viewer

kellyhaydon:
It would depend on the format, so having options makes sense.

retokromer:
What is the status of this? Is @savcampbell still working on vrecord?

retokromer:
Ping! Can this be merged?

dericed:
Adding self to assignment.

vrecord
```diff
@@ -1096,17 +1096,23 @@ if [[ "${standard_choice}" = "PAL" ]] ; then
...
     recordingfilter="[0:v:0]setfield=bff,setsar=40/27,setdar=4/3[vid1]"
 fi
 
+if [[ "${playbackview_choice}" = "Broadcast Range Visual" ]] ; then
+    middleoptions+=(-color_range mpeg)
+elif [[ "${playbackview_choice}" = "Full Range Visual" ]] ; then
+    middleoptions+=(-color_range jpeg)
+else
+    middleoptions+=(-color_range mpeg)
+fi
```
retokromer: Are these tests not equivalent to this one?
```
if [[ "${playbackview_choice}" = "Full Range Visual" ]] ; then
    middleoptions+=(-color_range jpeg)
else
    middleoptions+=(-color_range mpeg)
fi
```

retokromer:
@dericed I simplified the test (please check if I am not missing something), because @savcampbell seems no longer to be working on vrecord.

retokromer:
ping

dericed:
Is this tested? From a skim it seems that in some conditions that -color_range could be declared twice.

retokromer:
I am not sure I understand you carefully. The `-color_range` value is set only in https://github.com/savcampbell/vrecord/blob/0c94cb16340e61d54e946e4d6068e6dab79bd325/vrecord#L1108-L1112 which in the current version depends on one single test.

dericed:
ah ok i misread the patch.

PR #107, opened 2016-11-29
Gui for mipops

privatezero:
Adds secondary GUI to help with MIPoPS desktop icon based workflows. Also adds version variable based around homebrew install path and icon.


retokromer:
Not tested, but code LGTM.

dericed:
can you use a relative link for the png instead

dericed:
Accepting nonetheless as there's a lot about vrecord that would be difficult if not run through homebrew.

Issue #101, opened 2016-11-20
Documentation


retokromer:
Some thoughts on my side:
- Update the Manual if needed: it’s now covering 0.7.10, but the current release is 0.7.24 [if it’s not needed, I’ll just update the version number].
- Formally add a licence: I personally suggest CC BY 4.0, but I’m fine with others, like MIT.
- Does it make sense to separate into a short ReadMe (introduction on scope, how to communicate bugs and how to contribute) and the actual Manual (how to use the software)?
- Would an additional page about setting up the hardware be useful?

retokromer:
Includes https://github.com/amiaopensource/vrecord/issues/93 i.e. the Manual needs an update.

dericed:
I agree with setting README to CC-BY-V4. Ok with you, @jfarbowitz and @privatezero?

privatezero:
👍

jfarbowitz:
Sounds good to me. I'm also fine with putting the same license on the Analog Digitization How-To: https://github.com/amiaopensource/vrecord/blob/master/Resources/analog_digitization.md

retokromer:
OK with @savcampbell, I wait. May I add anther point to my wish list?

- Don’t put the prompt into the command, because it confuses beginners when they try to copy and paste the actual command: e.g. just write `brew install vrecord` instead of `$ brew install vrecord`.

savcampbell:
Removed the command prompts and added a license in #112

retokromer:
Thank you, @savcampbell! I guess more or less all is now solved in https://github.com/amiaopensource/vrecord/pull/112 and I close this one.

dericed:
Also closes https://github.com/amiaopensource/vrecord/issues/110

PR #117, opened 2016-12-05
add CC BY 4.0 licence

retokromer:
As suggested by @jfarbowitz (see https://github.com/amiaopensource/vrecord/issues/101).


dericed:
If ok with @jfarbowitz then ok with me.

Issue #124, opened 2016-12-18
Playback view menu not working properly


retokromer:
The `Which playback view?` does not accept `quit` value during the parametrisation process.

PR #125, opened 2016-12-18
correct "quit" option

retokromer:
Should resolve https://github.com/amiaopensource/vrecord/issues/124.


retokromer:
merged into https://github.com/amiaopensource/vrecord/pull/126/commits

PR #128, opened 2016-12-23
set transparency to 0.95

retokromer:
- set transparency to 0.95 for better visibility
- correct "quit" also in commented out part

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -541,9 +541,8 @@ if [[ "${runtype}" = "audiopassthrough" ]] ; then
 #[f1][abcde1]hstack[abcdef1],\
 #[abcdef1][xx]overlay=200:10[vo]"
 #            ;;
-#        Q) exit 0 ;;
-#        q) exit 0 ;;
-#        *)  report -w "Error: Not a valid option, select a valid number or [q] to quit." ; return 1 ;;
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -541,9 +541,8 @@ if [[ "${runtype}" = "audiopassthrough" ]] ; then
 #[f1][abcde1]hstack[abcdef1],\
 #[abcdef1][xx]overlay=200:10[vo]"
 #            ;;
-#        Q) exit 0 ;;
-#        q) exit 0 ;;
-#        *)  report -w "Error: Not a valid option, select a valid number or [q] to quit." ; return 1 ;;
+#        "quit") report -d "Bye then" ; exit 0 ;;
+#        *) report -w "Error: Not a valid option, select a valid number." ; return 1 ;;
```
dericed: report is called before the function is defined.

retokromer: report is defined at https://github.com/amiaopensource/vrecord/blob/f22cfc811b31a2ccc1adbe65c3c49c623d748715/vrecord#L253

dericed: ok, you're right, this doesn't impact this pr, though there's still a report on line 65 that won't work properly

retokromer: I’ll check that one (in another PR).

PR #129, opened 2016-12-23
order of functions’ call

retokromer:
1. usage
2. _get_iso8601
3. report

Should solve https://github.com/amiaopensource/vrecord/pull/128#discussion_r93799419


dericed:
Thx!

PR #132, opened 2017-01-06
add a Hardware page

retokromer:
Considering the current discussion on AMIA-L, a page about the needed hardware for the digitisation station would be useful.

cc @jfarbowitz

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,8 @@ man vrecord
 
 If you want to see a more detailed description about how to digitize analog videotape see our document on [analog digitization](Resources/analog_digitization.md).
 
+A short information about the [needed hardware](Resources/hardware.md) is provided as well.
+
 We want vrecord to be a helpful tool for audiovisual archivists and others. If you experience any problems with vrecord you can open a new issue with our Github [issue tracker](https://github.com/amiaopensource/vrecord/issues). Try to see if you can replicate the issue yourself first and describe in detail what factors led to it. Please let us know if you were able to succesfully replicate the issue. 
 
 Feel free to contribute to our github project by creating a fork and sending pull requests.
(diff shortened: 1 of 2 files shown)


dericed:
For vrecord I think the hardware needed is a computer running a recent Mac OS or Linux and Blackmagic card from:

- https://www.blackmagicdesign.com/products/ultrastudiothunderbolt
- https://www.blackmagicdesign.com/products/decklink
- https://www.blackmagicdesign.com/products/intensitypro4k
- https://www.blackmagicdesign.com/products/intensity

and whatever is necessary to connect that to the computer.

retokromer:
Thank you, @dericed. I’ll improve a little the placeholder, and cross check it also with my Albanian pals Skip and Jonathan.

jfarbowitz:
I've only ever seen vrecord run with with the Ultrastudio and the Decklink. Do we have confirmation that it will run on the Intensity?

Also, I know we tried Linux installations in the past and there were issues. Have those been resolved?

retokromer:
Did we not use the Intensity in Tirana?

There are still issues on Linux, but many parts seem to work now. And there are big issues on Windows and so far it almost doesn't work (also I would like to use Ubuntu on Windows 10). In South-East Asia macOS is hardly ever used, bu they have a huge amount of analogue cassettes to digitise...

TedLangdell:
According to Skip, it was an intensity shuttle thunderbolt addition on the $499 Mac Mini.

I've duplicated the same set up, and I'm using USB 3 Drive Successful Way to capture 8 bit YUV in an MOV wrapper using both the blackmagic  media express software and VRecord.

jfarbowitz:
@retokromer, my mistake. We did use the Intensity in Tirana.

retokromer:
No problem at all, @jfarbowitz, I was just unsure if I am mixing together different things. It happens…

PR #140, opened 2017-05-11
uniform syntax

retokromer:
- add `#` info at the beginning
- improve version in help message
- always use `_report` instead of `echo`
- uniform `exit` values
- delete not needed escaping of CR/LF
- alignment

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
+# vrecord
+# Open-source software for capturing a video signal and turning it into a digital file.
+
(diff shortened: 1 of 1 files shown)


privatezero:
I just took a look, and it looks like mm isn't actually listed as a dependency of vrecord?

retokromer:
Hmm… good point! This should be added, as a partial use of `mm` was already there.

retokromer:
No, I am wrong! The two functions of `mm` it uses (i.e. `_report` and `_get_iso8601`) are implemented directly into `vrecord`.

privatezero:
Oh nice!

PR #143, opened 2017-05-13
add brackets

retokromer:
- add brackets to variables
- alignement
- uniform PS3 with `: ` at the end of the message
- move `${version}` definition before it’s used
- use `[[ ... ]]` instead of `[ ... ]`


privatezero:
LGTM!

Issue #20, opened 2015-10-29
Prevent user from starting new capture when bmdcapture is still running


CSchloss385:
If bmdcapture is still running, a user can open up a new terminal tab or terminal window and start an entirely new capture. Need to find a way to tell the user that bmdcapture is busy and the user should close the terminal tab in order to proceed with a new capture.

jfarbowitz:
This might be helpful for writing to code: http://www.anyexample.com/linux_bsd/bash/check_if_program_is_running_with_bash_shell_script.xml

retokromer:
ping

retokromer:
@jfarbowitz Are you working on this?

PR #151, opened 2017-06-10
abort if `bmdcapture` is running

retokromer:
should resolve https://github.com/amiaopensource/vrecord/issues/20 (untested)

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1001,7 +1001,7 @@ else
 fi
 
 if ps ax | grep -v grep | grep 'bmdcapture' > /dev/null ; then
-    _report -w "bmdcapture is busy. You should close the terminal tab in order to proceed with a new capture."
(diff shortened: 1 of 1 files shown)


jfarbowitz:
The code looks good in theory to me, but I also can't test. @dericed are you able to test?

dericed:
I tested it and it works though I think the error message is misleading. For instance

```
./vrecord -p
bmdcapture is busy. You should close the terminal tab in order to proceed with a new capture.
```

When I then close that terminal tab and use another terminal window, I get the same error.

Suggested warning: `bmdcapture is already busy`

retokromer:
Thank you, @dericed! I updated the error message as suggested. I’ll merge this and your PRs tomorrow, and make a new release (for next week’s training in Bangkok).

PR #153, opened 2017-06-15
add bit plane viewer

dericed:
all for https://github.com/amiaopensource/vrecord/issues/138


kieranjol:
Looks great to me, very useful.

Out of curiosity, is there any significance to the colouring that shows up sometimes? i'd expected just black and white values. I see a lot more when I go full screen, far fewer in windowed:
screen shot 2017-06-15 at 10 13 08
screen shot 2017-06-15 at 10 09 32

dericed:
The bitplanes of all three planes are shown together. Y in b/w and U and V in colors. You can replicate it in qctools via the bitplane viewer by setting all planes to the same bit value.

kieranjol:
Ah, good to know, thanks Dave!

PR #160, opened 2017-07-22
test if FFmpeg is correctly installed

retokromer:
Prevents e.g. the error:
```
dyld: Library not loaded: /usr/local/opt/x265/lib/libx265.116.dylib
  Referenced from: /usr/local/bin/ffmpeg
  Reason: image not found
Abort trap: 6
```
This may occur e.g. when `x265` has been updated, after installing `ffmpeg --with-x265`.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -3,11 +3,9 @@
 # vrecord
 # Open-source software for capturing a video signal and turning it into a digital file.
 
-if ffmpeg -version 2>&1 | grep "Library not loaded" >/dev/null; then
-    echo "Please wait while reinstalling 'FFmpeg'..."
-    brew reinstall ffmpeg \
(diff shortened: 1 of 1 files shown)


dericed:
The approach seems a little intrusive. Would prefer to warn and advise rather than start installing. For instance on some machines I install vrecord under admin user and it is run under a non-admin user, this scenario would cause subsequent install errors.

retokromer:
We got this very error last week in Salemi, Sicily. Therefore I had the idea to fix it, because it may be tricky to understand by the user. Yet I will amend by only «translating» the system error message in a (hopefully) less cryptic one.

In my company we do install as admin and then work as non-admin user. A similar code did not cause a subsequent install error.

retokromer:
@dericed Is this version fine with you?

Issue #169, opened 2017-08-08
Broadcast Range Pixels not highlighting


iamdamosuzuki:
Using version 2017-07-27

Full Range pixels are highlighting properly, but Broadcast Range pixels are not. 

![screen shot 2017-08-08 at 4 03 04 pm](https://user-images.githubusercontent.com/3967767/29098438-3ace1d3e-7c54-11e7-81fd-15b29724caab.png)
![screen shot 2017-08-08 at 4 03 18 pm](https://user-images.githubusercontent.com/3967767/29098440-3cf3bbc8-7c54-11e7-8e4f-d58a25863c1d.png)

retokromer:
@iamdamosuzuki By chance, can you name me correct working version? The last two releases were «only» documentations improvements and a test if FFmpeg is correctly installed… Otherwise I will delve into the commit.

iamdamosuzuki:
It's working fine on this version:

`vrecord 0.7.28`

Here's a screenshot

![screen shot 2017-08-11 at 5 07 45 pm](https://user-images.githubusercontent.com/3967767/29235822-8b83ee06-7eb7-11e7-8f41-d92a74c0d11a.png)

retokromer:
@iamdamosuzuki Thank you!

retokromer:
Hmm…

I cannot test with a setting those days, but reading the code, the only [differences](https://github.com/amiaopensource/vrecord/compare/v0.7.28...v2017-07-27) I can see between `0.7.28` and `2017-07-27` (or the current `HEAD`) are:

- add [`-color_range jpeg`](https://github.com/amiaopensource/vrecord/commit/f84e75950ae5422e497eba2e3c80ddcb38bcba17) for full range, instead of `-color_range mpeg` in all the other cases;
- modify the [playbackfilter](https://github.com/amiaopensource/vrecord/commit/99df0d65a3b37798a3ebf7483a1ea1eecb3c2682) for the `Visual + Numerical` case;
- add a [playbackfilter](https://github.com/amiaopensource/vrecord/commit/1e52a68c4e7960ed41e63718bd73fc71667c04bc) for the new `Bit Planes` case.

Yet no one of these seems to me related to this bug… or am I missing something? Any thoughts @dericed @privatezero and/or @savcampbell? Thanks!

dericed:
[...]

PR #172, opened 2017-08-13
add test on broadcast range

retokromer:
- trying to fix https://github.com/amiaopensource/vrecord/issues/169


retokromer:
Yes, @dericed, that was **_exactly_** the reason why I changed Savannah Campbell’s original code (which is the one in this PR). Thank you!

retokromer:
Where is the bug still remains obscure… to me, sorry.

PR #175, opened 2017-08-23
add caveat about Mini Converter

retokromer:
might be clearer, considering the [discussion](https://github.com/amiaopensource/vrecord/issues/169#issuecomment-324116602) on the Mini Converter

### DIFF
--- a/Resources/hardware.md
+++ b/Resources/hardware.md
@@ -1,4 +1,4 @@
-# Needed Hardware #
+# Needed Hardware
 
 - a computer running a recent macOS or Linux
 - a Blackmagic card from:
@@ -7,9 +7,11 @@
   - [DeckLink](https://www.blackmagicdesign.com/products/decklink)
   - [Intensity Pro 4K](https://www.blackmagicdesign.com/products/intensitypro4k)
   - [Intensity](https://www.blackmagicdesign.com/products/intensity)
+  
+  Please note that the Analog to SDI Mini Converter does not work well, because the signal is clipped at broadcast levels.
 
 - the necessary [cables](https://amiaopensource.github.io/cable-bible/) for connecting the video desk to the card and to the computer
 
-## License ###
+## License
 
 <a rel="license" href="https://creativecommons.org/licenses/by/4.0/"><img alt="Creative Commons License" style="border-width:0" src="https://i.creativecommons.org/l/by/4.0/80x15.png"></a><br>Vrecord is licensed under a <a rel="license" href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0 International License</a>.
--- a/Resources/hardware.md
+++ b/Resources/hardware.md
@@ -1,6 +1,6 @@
 # Needed Hardware
 
-- a computer running a recent macOS or Linux
+- a computer running a recent macOS
 - a Blackmagic card from:
 
   - [UltraStudio](https://www.blackmagicdesign.com/products/ultrastudiothunderbolt)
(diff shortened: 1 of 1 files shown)


kieranjol:
I think we might have that same box, and it has hardware switches for
levels, signal type etc. Perhaps this could be related to the signal
clipping? @iamdamosuzuki

kieranjol:
No, I could try it out and mess with the switches though. Hopefully morgan
can chime in on the hardware settings. We have been using digibeta decks
for betacam sp migration as we were getting some signal interference with
this convertor + our betasp deck. It was probably just a cabling /patch
bay/user issue though, I doubt the convertor had anything to do with it.

iamdamosuzuki:
I spent the morning messing around with the dip switches and couldn't get the box to not clip. Here's the configuration I used:

1 -> OFF -> Processing On
2 -> OFF -> Blank
3 -> OFF -> Blank
4 -> OFF -> Component (using the component out of a DPS - 295)
5 -> OFF -> N/A (switch 5 only matters if switch 4 is ON)
6 -> OFF -> SMPTE Levels
7 -> OFF -> 7.5 IRE
8 -> OFF -> Analog Audio

So this configuration clips in the Mini Converter box, and I couldn't find another configuration that doesn't clip. @kieranjol, do you think you'll have a chance to try sending the box some high levels and seeing if it clips at 100 IRE? Also, Kathy O'Regan says hi!

kieranjol:
Wow, that's some nice research there Morgan! I'll do my best to try out vrecord tomorrow with that converter.
@retokromer , this converter just receives an analog signal and outputs SDI, you'd need additional Blackmagic hardware to capture the SDI. Though there are some really cheap Blackmagic devices that do not do any analog to digital conversion, but they would receive SDI, so they could be used in tandem with this device I guess.

retokromer:
@kieranjol If the SDI signal is always clipped by this device, then I thing it’s not suitable for our purpose.

kieranjol:
agreed

retokromer:
In my understanding, the Mini Converters are not suitable for vrecord when full range is wished. Therefore I merge this caveat in the documentation. Please feel free to amend.

kieranjol:
@iamdamosuzuki So this is some bars from a BetaSP tape using the same dip switch settings as you. Vrecord is showing values outside of broadcast range when using the default Visual +numerical view, an the same when explicitly selecting 'broadcast range visual'.

My setup here is a Sony UVW-1800P ->component out to Blackmagic Mini Converter Analog to SDI, which is passed via SDI to Blackmagic Ultrastudio 4K into vrecord 2017-07-27 using ffmpeg 3.3.3. I don't have a TBC hooked up, I knew the signal on this tape was quite hot to begin with so I just left it run with default levels.
screen shot 2017-08-24 at 09 41 04

kieranjol:
I'd imagine that this converter has whatever default firmware it shipped with. I don't think we ever hooked it up to a computer to upgrade it or anything. Is there any way that some other factor in the signal chain is causing your clipping rather than the mini converter?

iamdamosuzuki:
@kieranjol Very interesting. The last thing I can think of is that the UltraStudio Black Magic I/O I'm using is clipping the SDI input. I was able to get levels over broadcast range using the analog inputs from the breakout cable, but I didn't consider that the SDI input might be where the problem is. I'll do some more testing next week and report back, thanks so much for helping out!

kieranjol:
I look forward to hearing more. Also say hi to Kathy for me, Up Gaillimh etc..

iamdamosuzuki:
I haven't had any luck with the Mini Converter. No matter how many configurations I put the Dip Switches in it still clips the signal, though I haven't tried upgrading the firmware. When I replaced the Mini Converter with a DPS 470 as an A/D I was able to get full range without clipping. This was using the same BlackMagic UltraStudio HD box as an I/O. It's very strange that you're getting different results with the same box!

PR #179, opened 2017-08-29
"full range visual" fix

MIAPtech:
fix "Full Range Visual" to highlight pixels at full range extremes rather than at broadcast limits; re-worded description in README

addresses issue #173 

all other diffs are trailing whitespaces that I stripped out by accident, as I didn't realize Atom does that by default. can re-submit but I don't think it hurts anything?

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ All of options in the vrecord GUI (which appears when running `vrecord -e`) or o
 
 **Broadcast Range Visual mode** — Broadcast Range Visual mode displays the video feed, the video feed with pixels out of broadcast range highlighted, the waveform monitor, and the vectorscope in the vrecord window.
 
-**Full Range Visual mode** - Full Range Visual mode displays the video feed, the video feed with pixels out of full range highlighted, the waveform monitor, and the vectorscope in the vrecord window.
+**Full Range Visual mode** - Full Range Visual mode displays the video feed, the video feed with pixels at full range extremes highlighted, the waveform monitor, and the vectorscope in the vrecord window.
 
 **Visual + Numerical mode** — Visual and Numerical mode displays your favorite video feeds and scopes as well as numerical values for the characteristics of the video signal in the left sidebar. 
 
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ This documentation is up to date as of vrecord version 2017-07-15.
 
 ## Summary
 
-Vrecord is open-source software for capturing a video signal and turning it into a digital file. Its purpose is to make videotape digitization or transfer easier. Vrecord can capture analog and digital signals through a variety of inputs and can create digital video files in a variety of formats and codecs. Vrecord has been designed with needs of audiovisual archivists in mind.
+Vrecord is open-source software for capturing a video signal and turning it into a digital file. Its purpose is to make videotape digitization or transfer easier. Vrecord can capture analog and digital signals through a variety of inputs and can create digital video files in a variety of formats and codecs. Vrecord has been designed with needs of audiovisual archivists in mind. 
 
 Vrecord uses [ffmpeg](http://ffmpeg.org), [ffplay](http://ffmpeg.org/ffplay.html), and [bmdcapture](https://github.com/lu-zero/bmdtools) to do its dirty work.
 
-Currently vrecord only supports Blackmagic Design [capture cards](Resources/hardware.md) with the Blackmagic driver installed.
+Currently vrecord only supports Blackmagic Design [capture cards](Resources/hardware.md) with the Blackmagic driver installed. 
 
 ## Installing vrecord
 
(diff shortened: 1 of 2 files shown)


vrecord
```diff
@@ -579,7 +579,7 @@ split=5[a][b][c][d][e];\
 [c1]${WAVEFORM_FILTER}[c2];\
 [a][b2][c2]vstack=inputs=3,format=yuv422p[abc1];\
 [d]${VECTORSCOPE_FILTER}[d1];\
-[e]format=yuv444p,lutyuv=y=if(eq(val\,maxval)\,minval\,if(eq(val\,minval)\,maxval\,val)):u=val:v=val,scale=512:ih[e1];\
+[e]format=yuv444p,lutyuv=y=if(eq(val\,254)\,1\,if(eq(val\,1)\,254\,val)):u=val:v=val,scale=512:ih[e1];\
```
dericed: Why only consider the value of `254`? Perhaps it should be `ge` rather than `eq`.

README.md
```diff
@@ -26,11 +26,11 @@ This documentation is up to date as of vrecord version 2017-07-15.
 
 ## Summary
 
-Vrecord is open-source software for capturing a video signal and turning it into a digital file. Its purpose is to make videotape digitization or transfer easier. Vrecord can capture analog and digital signals through a variety of inputs and can create digital video files in a variety of formats and codecs. Vrecord has been designed with needs of audiovisual archivists in mind. 
+Vrecord is open-source software for capturing a video signal and turning it into a digital file. Its purpose is to make videotape digitization or transfer easier. Vrecord can capture analog and digital signals through a variety of inputs and can create digital video files in a variety of formats and codecs. Vrecord has been designed with needs of audiovisual archivists in mind.
```
dericed: nitpick, i'd suggest to having spacing changes in separate commits than content changes.

MIAPtech: sorry, have been flustered trying to get Atom settings to stop automatically eliminating whitespace. new commit without spacing changes below

vrecord
MIAPtech: Does `ge` calculate a "greater than or equal to [x] value" rather than just "is equal to [x]"?

That's logical and I can test tomorrow or Thurs. I set the thresholds and 1 and 254 because 0 and 255 didn't display any highlighting - assumed that values were getting clipped (by the Blackmagic hardware?) right at 0 and 255 and thus the filter wasn't reading/highlighting those values correctly.

Are you saying that pixels could have values at 255 in vrecord and get captured at those too-high levels without getting clipped? At what point does a pixel with "objective" value at 255 or over get clipped?

dericed: the expressions are listed at http://ffmpeg.org/ffmpeg-utils.html#Expression-Evaluation.  From a skim, oops, I guess I meant `gte()` not `ge()`.

I'd suggest `gte(val\,254)\,minval\,if(lte(val\,1)\,maxval\,val))`, but I'm unclear the intent, as in which values exactly should be highlighted.

MIAPtech: well figuring out the intent of the feature was part of what I was trying to get at in my question in #173, because maybe I'm still just not understanding "full range" - but I don't understand how pixels can be "out of" full range values, as far as vrecord goes. They are clipped at 0 or 255, no?

So I would think pixels *at* those values should be highlighted to signal the operator that information may be getting lost in the capture and they may want to reign in luminance. My suggestion was just based on replacing "minval" and "maxval" variables where I saw them, since they seem to correspond to broadcast range. I very much admit I don't get everything happening in this line!

Does that help at all? I can test your suggested line tomorrow!

MIAPtech: OK,
`gte(val\,254)\,minval\,if(lte(val\,1)\,maxval\,val))` and
`gte(val\,255)\,minval\,if(lte(val\,1)\,maxval\,val))`

display essentially the same (desired) highlighting at extremes for me (using `0` instead of `1` still displays no highlighting at all when signal hits the low end)

I think the latter best matches intent and using `gte` and `lte` covers out-of-range scenarios others may see? (maybe my hardware is clipping in a specific manner others don't observe)

retokromer: If I remember carefully, in 8 bit the values 0 and 255 should not be allowed, because they are reserved for synchronisation.

dericed: That is true in some context. For instance Apple's TN2162 reserves those values in uncompressed YCbCr in QuickTime, but those reservations aren't true in all context; a value of 255 in uncompressed in MKV is no big.

retokromer: Therefore, if I understand the filter correctly, 1 and 254 would be desired.

retokromer:
FYI (no rush): If this is resolved soon, then I will made a new release for the next teachings.

vrecord
retokromer: @dericed Do you mean that the filter should act differently, according to the chosen container/codec pair?

retokromer:
@dericed Can I merge this?

dericed:
Fine with me, though gte/lte at 1/254 would seem safer to me than eq 1/254.

retokromer:
@MIAPtech Is this fine also with you?

MIAPtech:
Agreed, `gte()` and `lte()` make more sense!

retokromer:
Thank you very much, @MIAPtech!

PR #185, opened 2017-10-11
adds less specific qctools_xml conditional check

kieranjol:
The capture process will only launch if `$qctoolsxml_choice` is set to `Yes` or `No`.
If it is set to `undeclared` in the conf, then the terminal interview will appear.

By default, the value of `$qctoolsxml_choice` is `''`, so the interview does not launch. This pull request causes the capture process to run without a qctools option if `$qctoolsxml_choice` is set to anything but `Yes`.

Another way around this would be to make the terminal interview appear if `$qctoolsxml_choice` is set to `''`, but perhaps there is a reason why this is not the current behaviour?


kieranjol:
To be more specific, if your config file is empty and you do not specify an option in the GUI or in the config file, vrecord will not launch the capture process. You have to select Yes/No/Undeclared in order to launch the capture process.

PR #188, opened 2017-10-16
use qcli for qctools reports

dericed:


dericed:
could resolve https://github.com/amiaopensource/vrecord/issues/155

privatezero:
Can't test right now, but looks good to me!

Would then have to update https://github.com/amiaopensource/homebrew-amiaos/blob/master/vrecord to require qcli then, right?

dericed:
I'll make a release after the merge and then update the dependencies.

vrecord
```diff
@@ -1193,8 +1193,12 @@ else
 fi
 
 if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
+    if [[ ! "$(which qclfi)" ]] ; then
```
kieranjol: Is that f supposed to be there

dericed:
Oops.

retokromer:
What about the possibility to give to the user the chose between `qcli` (with an additional dependence) and `ffprobe` (without an additional dependence)?

dericed:
updated for typo. to @retokromer both ffprobe and qcli are an additional dependence, this is the only place where ffprobe is used.

retokromer:
@dericed As long as I remember, on brew ffprobe comes by default with ffmpeg.

PR #190, opened 2017-10-17
move from bmdcapture to ffmpeg

dericed:
requires `brew install ffmpegdecklink --HEAD`

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1132,7 +1132,7 @@ else
     done
 fi
 
-_report -d "Summary: ${pixel_format} bit ${codecname} ${container_choice} file from ${standard_choice} ${video_input_choice} ${audio_input_choice} Frame MD5s=${framemd5_choice} QCTools XML=${qctoolsxml_choice} and Technician=${technician}. Inputs recorded to ${dir}/${id}.${extension} Auxiliary Files created in ${logdir}"
+_report -d "Summary: ${pixel_format} bit ${codecname} ${container_choice} file from ${standard_choice} ${video_input_choice} ${audio_input_choice} Frame MD5s=${framemd5_choice} QCTools XML=${qctoolsxml_choice} and Technician=${technician}. Inputs recorded to ${dir}/${id}${suffix}.${extension} Auxiliary Files created in ${logdir}"
 if [[ "$invert_phase" = 1 ]] ; then
     echo -e " \033[101mWARNING: Option to invert phase of second audio channel has been selected\033[0m"
 fi
--- a/vrecord
+++ b/vrecord
@@ -1115,6 +1115,12 @@ else
     done
 fi
 
+if [ -f "${dir}/${id}${suffix}.${extension}" ] ; then
+    _report -w "A file called ${dir}/${id}.${extension} already exists."
+    _report -w "Exiting to avoid overwriting that file."
+    exit
+fi
+
 if [[ "${audio_mapping_choice}" && "${audio_mapping_choice}" != "${undeclaredoption}" ]] ; then
     lookup_audio_mapping "${audio_mapping_choice}"
 else
--- a/vrecord
+++ b/vrecord
@@ -1191,7 +1191,6 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
         _report -w "Such as \`brew install qcli\`."
         exit 1
     fi
-    QCTOOLS_PRE=(-map 0 -c copy -f nut -)
     QCTOOLS_POST=(qcli -i - -o "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz")
 else
     QCTOOLS_POST=(cat)
@@ -1201,16 +1200,16 @@ fi
     -v info -hide_banner -nostats "${inputoptions[@]}" "${GRAB_DECKLINK[@]}" \
     -metadata:s:v:0 encoder="${codecname}" "${middleoptions[@]}" -c:a pcm_s24le \
     -filter_complex "\
-        [0:v:0]split[record0][play0];\
-        [record0]${recordingfilter}[record1];\
-        [play0]${playbackfilter},format=yuv420p[play1];\
+        [0:v:0]${recordingfilter};\
         ${audiomap}" \
     $(if [ -n "$map1k" ] ; then echo ${map1k} ; fi) $(if [ -n "$map1v" ] ; then echo ${map1v} ; fi) \
     $(if [ -n "$map2k" ] ; then echo ${map2k} ; fi) $(if [ -n "$map2v" ] ; then echo ${map2v} ; fi) \
-    -map "[record1]" -f "${format}" "${dir}/${id}${suffix}.${extension}" \
-    -map "[play1]" -window_title "mode:${runtype} - video:'${video_input_choice}' audio:'${audio_input_choice}'" \
-    -f sdl sdl_dummy_output \
-    "${extraoutputs[@]}" "${QCTOOLS_PRE[@]}" | \
+    -f "${format}" "${dir}/${id}${suffix}.${extension}" \
+    "${extraoutputs[@]}" -map 0 -c copy -f nut - | \
+    tee >("${FFPLAY_DECKLINK[@]}" \
+        -v info -hide_banner -stats -autoexit -i - \
+        -window_title "mode:${runtype} - video:'${video_input_choice}' audio:'${audio_input_choice}'" \
+        -vf "${playbackfilter}") | \
     "${QCTOOLS_POST[@]}"
 
 if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
--- a/vrecord
+++ b/vrecord
@@ -1197,7 +1197,7 @@ else
 fi
     "${FFMPEG_DECKLINK[@]}" \
     2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
-    -v info -hide_banner -nostats "${inputoptions[@]}" "${GRAB_DECKLINK[@]}" \
+    -v info -nostdin -hide_banner -nostats "${inputoptions[@]}" "${GRAB_DECKLINK[@]}" \
     -metadata:s:v:0 encoder="${codecname}" "${middleoptions[@]}" -c:a pcm_s24le \
     -filter_complex "\
         [0:v:0]${recordingfilter};\
--- a/vrecord
+++ b/vrecord
@@ -3,11 +3,6 @@
 # vrecord
 # Open-source software for capturing a video signal and turning it into a digital file.
 
-if ffmpeg -version 2>&1 | grep "Library not loaded" >/dev/null ; then
-    echo "Please reinstall 'FFmpeg'. Exiting."
-    exit 1
-fi
-
 config_file="$HOME/.$(basename "${0}").conf"
 if [[ $(dirname "$(which "${0}")") = "/usr/local/bin" ]] ; then
     version=$(TMP=$(brew info vrecord | grep -Eo "/vrecord/.* \(") ; echo ${TMP:9:(${#TMP}-11)})
@@ -21,7 +16,7 @@ unset suffix
 unset duration
 unset technician
 runtype="record"
-bmdcapturelogsuffix="_bmdcapture.log"
+capturelogsuffix="_ffmpeg_decklink_input.log"
 ffmpeglogsuffix=".log"
 unset pashuainstall
 unset extraoutputs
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -439,10 +431,10 @@ lookup_container(){
 lookup_standard(){
     case "${1}" in
         "NTSC")
-            standard="${NTSC_value}"
+            standard="ntsc"
             ;;
         "PAL")
-            standard="${PAL_value}"
+            standard="pal "
```
retokromer: Is the space in `"pal "` a typo?

retokromer:
BTW: `ffmpegdecklink` could be updated to `ffmpeg 3.4`.

vrecord
```diff
@@ -957,25 +946,28 @@ else
...
-GRAB_DECKLINK+=(-F nut)
-GRAB_DECKLINK+=(-f pipe:1)
+GRAB_DECKLINK=(ffmpeg-dl)
+GRAB_DECKLINK+=(-f decklink)
+GRAB_DECKLINK+=(-audio_input "${audio_input}")
+GRAB_DECKLINK+=(-video_input "${video_input}")
+GRAB_DECKLINK+=(-format_code "${standard}")
+GRAB_DECKLINK+=(-channels 8)
+GRAB_DECKLINK+=(-raw_format "${pixel_format}")
+GRAB_DECKLINK+=(-i 'UltraStudio Express')
```
kieranjol: Is this a hardcoded value?what if you had another decklink or ultrqstudio card?

dericed: I didn’t get around to writing a query for the devices yet.

vrecord
```diff
@@ -1142,7 +1134,7 @@ _writeingestlog "computer_name" "$(uname -n)"
 _writeingestlog "user_name" "$(whoami)"
 _writeingestlog "operating_system_version" "$(uname -v)"
 _writeingestlog "datetime_start" "$(_get_iso8601)"
-_writeingestlog "video_bit_depth" "${video_bitdepth}"
+_writeingestlog "video_bit_depth" "${pixel_format}"
```
retokromer: Change also the labelling from `video_bit_depth` to `pixel_format `?

vrecord
kieranjol: cool

retokromer:
Q: Is there a reason for choosing `brew install ffmpegdecklink --HEAD`? Does the last release and `brew install ffmpegdecklink` not work?

dericed:
@retokromer ffmpegdecklink --HEAD would give access to https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/89cc48551bbe9f147ba9f4ca3821a35797cf7b47, which I actually haven't yet used in this patch, but intend to

retokromer:
@dericed Thank you very much for your answer! I understand. The downside I see with the HEAD is that it’s more difficult to manage in the field.

dericed:
@retokromer hopefully we can improve on this issue before this is merged.

vrecord
dericed: Hmm, the issue here is whether the use picks between `8` and `10` or between `yuv422p10` or `uyvy422`.

privatezero:
I am curious about the effects of this switch on how the capture device handles different line levels for the audio inputs.  I notice that there are different options for setting `audio_input` to either `analog_xlr` or `analog_rca.`

I believe that the UltraStudio is set by default to expect a 'professional' line level (+4 dBu) and that it takes a setting in its GUI to switch it over to the consumer line level (-10 dBv). Do we know A: what it is doing for expected level just leaving the setting on `analog`, and B: if using the more specific commands is it making the changes for expected level?

This also makes me wonder about the current levels in using bmdcapture! In looking at some of the bmdcapture info, it looks like setting the device to `-A 1` is setting it to RCA, which might imply expecting inputs at the consumer level?

privatezero:
When I emailed Blackmagic about input levels to the UltraStudio here is their response. If I recall correctly +4 dBm is the same as +4 dBu. Also Blackmagic documentation for `Use HiFi Audio Levels` says to apply that setting when using consumer RCA equipment, so even though in this email they said they don't support -10 dBV, it looks like this is in fact what the two settings mean.

> Thank you for contacting us. The UltraStudio Express does not have a setting to adjust the reference level for analog audio between +4dBu or -10dBv. The only setting it has that adjusts the reference level is labelled "Use HiFi Audio Levels," which adjusts the reference for RCA sources that use the HiFi format. For all other audio sources, there is only one audio reference level, which is +4dBm.

vrecord
retokromer: Indeed, yet in my opinion – not blocking at all! – it does not hurt if called what is actually is.

dericed:
Updated. With `brew install amiaopensource/amiaos/decklinkffmpeg --with-sdl` the playback and recording will work

vrecord
```diff
@@ -33,14 +28,19 @@ BRNG_OUTLIER_THRSHLD=14
 FFMPEG_DECKLINK=(ffmpeg-dl)
 FFPLAY_DECKLINK=(ffplay-dl)
 
+if "${FFMPEG_DECKLINK[@]}" -version 2>&1 | grep "Library not loaded" >/dev/null ; then
+    echo "Please reinstall 'ffmpeg-dl'. brew install amiaopensource/amiaos/ffmpegdecklink Exiting."
```
retokromer: I would suggest something like:
```
    echo "Please reinstall 'ffmpeg-dl':"
    echo "    brew install amiaopensource/amiaos/ffmpegdecklink"
    echo "Exiting."
```
I guess, it’s more understandable. (Or with `cat`.)

vrecord
```diff
@@ -33,14 +28,19 @@ BRNG_OUTLIER_THRSHLD=14
...
 _usage(){
     cat <<EOF
 $(basename "${0}") ${version}
 
-$(basename "${0}") will record a file via the Blackmagic SDK and bmdtools. It is an
+$(basename "${0}") will record a file via the Blackmagic SDK and ffmpeg. It is an
 interactive script and will create 8 or 10-bit video files.
 
-Dependencies: cowsay, ffmpeg, ffplay, mpv, qcli, and xmlstarlet
+Dependencies: cowsay, amiaopensource/amiaos/ffmpegdecklink --with-sdl2, mpv, qcli, and xmlstarlet
```
retokromer: I suggest:
```
Dependencies: cowsay, amiaopensource/amiaos/ffmpegdecklink --with-sdl2, mpv,
  qcli and xmlstarlet
```
in order to fit into 80 chars.

retokromer:
Can the tiny conflict be resolved, and then this be merged?

vrecord
```diff
@@ -3,42 +3,44 @@
...
 undeclaredoption="Undeclared"
 defaultfont="/Library/Fonts/Andale Mono.ttf"
 SAT_OUTLIER_THRSHLD=14
 AUD_OUTLIER_THRSHLD=10
 BRNG_OUTLIER_THRSHLD=14
+FFMPEG_DECKLINK=(ffmpeg-dl)
+FFPLAY_DECKLINK=(ffplay-dl)
+
+if "${FFMPEG_DECKLINK[@]}" -version 2>&1 | grep "Library not loaded" >/dev/null ; then
+    echo "Please reinstall 'ffmpeg-dl'. brew install amiaopensource/amiaos/ffmpegdecklink Exiting."
```
retokromer: I would replace also here `'ffmpeg-dl'`.

dericed:
This is ready for new testing, run:
`brew reinstall ffmpeg-decklink --with-sdl2 --with-freetype` first.

bturkus:
dying to try this, but could someone provide full instructions for those of us who are github-challenged? does something need to be cloned or pulled or fetched? And for those who might be wondering, I already attended @dinahhandel and @EG-tech's github workshop :)

what comes after this? 
brew reinstall ffmpeg-decklink --with-sdl2 --with-freetype

privatezero:
@bturkus, One way to do it would be to open a terminal window and run

`git clone https://github.com/amiaopensource/vrecord.git -b ffmpeg-decklink-WIP`

(Maybe first `cd` somewhere separate from your usual repos like your desktop to avoid getting it mixed up).

using the `-b` flag allows you to clone a specific remote branch.  Then you could just `cd` into that directory and run the new version of vrecord. Make sure to run it with `./vecord` instead of `vrecord` to run the one contained in the directory rather than the vrecord installed to your system path!

bturkus:
After updating to desktop video 10.9.7, vrecord's now working with the blackmagic decklink duo 2 card. thanks to everyone for making this happen! (still not working with vrecord classic, aka bmdcapture, but super pumped nonetheless).

I am noticing a few problems though, and wondering if others are seeing the same:
(1) After about 90 seconds of capturing to both ffv1/mkv and v210/mov, we're getting an error message that reads: "Decklink input buffer overrun!" If we continue much past this point, the resulting video has major audio sync issues, like the audio drops out completely after 2 min. We also get cowsay "pts discontinuities" errors for many, many frames (I can share an output if needed, but it's a lot).
(2) It almost seems as if vrecord (when in capture mode) isn't capturing in realtime, and the gap between what's being played and what's being captured gets greater and greater until the buffer gets overrun. I'm not seeing a similar delay when in passthrough mode.
(3) Almost every test file is generating a "clipped audio levels" cowsay error. Not sure if this is related to the above, but thought it might be worth mentioning. Based upon @privatezero's earlier comments, I did compare vrecord and media express audio levels; all seemed good.

thanks again!

kieranjol:
I've seen number to happen in an experimental build where the CPU could not
keep up. I think I was simultaneously capturing prores_ks and ffv1. I'd be
curious to know if your CPU or maybe your storage can't keep up (I doubt
the latter unless using usb2 or something)

retokromer:
@bturkus : Try a
```
sudo purge
```
before running vrecord.

dericed:
I replicated what @bturkus was seeing and adding a new commit. It reorders the ffmpeg outputs to have the output file first and sdl output (playback window) second. Also I rewrote it to use one filter_complex with split rather than as before where there was a mix of filter_complex and vf in the same command.

retokromer:
I guess, the change to filter_complex will make a difference.

vrecord
```diff
@@ -1191,46 +1191,37 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
...
-            -map "[vid1]" \
-            $(if [ -n "$map1k" ] ; then echo ${map1k} ; fi) $(if [ -n "$map1v" ] ; then echo ${map1v} ; fi) \
-            $(if [ -n "$map2k" ] ; then echo ${map2k} ; fi) $(if [ -n "$map2v" ] ; then echo ${map2v} ; fi) \
-            -f "${format}" \
-            "${dir}/${id}${suffix}.${extension}" \
-            "${extraoutputs[@]}"
+    QCTOOLS_POST=(cat)
 fi
+    "${FFMPEG_DECKLINK[@]}" \
+    2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
```
retokromer: I would indent while using `\`, but possibly it’s just… my personal perversion!

privatezero:
Having a buffer over run sounds like a similar problem to what I ran into setting up audiorecorder! Ultimately that also was solved be reordering the output file to come before the visualization pipe.

dericed:
Patch updated to clean up the commit log and make it a bit easier to review.

dericed:
The performance seems better, but there's a few pending issues. Currently one ffmpeg command is doing everything (read from decklink input, writes file, and plays video via filters in sdl); however `-f sdl` seems to not support audio. Anyone know how to get audio to play in this same command? Otherwise I think we have to return to piping from ffmpeg back to ffplay as before.

privatezero:
It looks like [pulseaudio](https://github.com/Homebrew/homebrew-core/blob/master/Formula/pulseaudio.rb) can be installed via homebrew so possibly you could get the `pulse` output device working?

Then potentially you could do something like `-map 0:v:0 -f sdl - -map 0:a:0 -f pulse -`?

No idea if this would work, and I haven't been able to test it, but a similar command using `alsa` seems promising (can't fully test as I seem to have messed up my sdl)

privatezero:
ok, now can confirm that (on linux because of alsa) `ffmpeg -i templeofthedog.mp4 -i map 0:v:0 -f sdl - -map 0:a:0 -f alsa default` allows simultaneous sound output while using sdl (just have to close both devices to exit).

If `pulse` could be got working, or there was another viable osx friendly audio out device this could be a strategy!

privatezero:
MORE continuation (sorry for the lack of brevity) I can confirm that running `brew install Pulseaudio` and then altering the install formula of ffmpeg to allow `--enable-libpulse`allows the simultaneous output of audio and sdl video via `ffmpeg -i INPUT.mp4 -map 0:a:0 -f pulse - -map 0:v:0 -f sdl TESTWINDOW`.

This still has sync issues, and I am still testing on a linux machine, but since this is using brew formulas it gives me hope this might work on macOS.

dericed:
thx Andrew. I'll test more tomorrow though today I was finding that:
`ffmpeg ... -f sdl ...`
would drop frames whereas
`fffmpeg ... -c copy -f nut - | ffplay -i -`
would be stable.

vrecord
```diff
@@ -1115,6 +1115,12 @@ else
     done
 fi
 
+if [ -f "${dir}/${id}${suffix}.${extension}" ] ; then
+    _report -w "A file called ${dir}/${id}.${extension} already exists."
+    _report -w "Exiting to avoid overwriting that file."
+    exit
+fi
```
retokromer: 👍

GregH18:
Hi all,
Thanks for making vrecord and constantly trying to improve it. We were running some tests with it and came upon some issues that I'm not 100% sure whether they've been raised yet or if anyone else is having them. 

1) When running `vrecord -a` audio can be heard through headphones in pass-through mode but not in record mode, and a message appears at the bottom of the window that says "Unsupported number of audio channels" and also "8 channels (48000 hz)" and the error in the image below. 
image01

2) We ran several tests using a combination of containers and codecs and we set the “set recording time in minutes” drop down to “0” and the video and logs were captured. However, when we started the capture, the line at the bottom that displays recording information while recording was not scrolling (see the image below). When the "set recording time" drop down is set to something other than "0", the ffplay window does appear. It might be helpful to have an error message or warning when setting the “set recording time in minutes” to 0. 
image02

3) QCTools reports cannot be created because it says “No such file or directory” (regardless of the container settings or what the capture duration is) but ffmpeg and ffplay reports/logs have been created, and the below the pop-up error message occurs sometimes (this probably partially due to us setting the record time to "0"). However, a video file is still captured. 
image03

4) When we capture using the following settings, we receive the error message below: 
image04

I hope this is helpful in working out the bugs and that it is not too confusing. 

Thanks again!

dericed:
I've tested a day of captures with this and thinking it's ready to merge and then make a release and update the vrecord installer. Any other comments? psst @bturkus

retokromer:
We also tested, and I think it’s OK to merge.

dericed:
hi @GregH18, could you break up these issues into separate items in the issues section at https://github.com/amiaopensource/vrecord/issues

retokromer:
>thx Andrew. I'll test more tomorrow though today I was finding that:
ffmpeg ... -f sdl ...
would drop frames whereas
fffmpeg ... -c copy -f nut - | ffplay -i -
would be stable. You can see the diff by using the vrecord in this branch and trying with and without the last commit dropped.

FYI @dericed I made good experiences with nut too, used in very different contexts.

Issue #193, opened 2017-10-27
Matrix Coefficients different for PAL Matroska vs Quicktime


beccabender:
When creating PAL FFV1 files in vrecord, the matrix coefficients value reported by mediainfo is different depending on whether the container is matroska or quicktime. 

Matroska:
``      <matrix_coefficients>BT.470 System B, BT.470 System G</matrix_coefficients>
``

Quicktime:
``   <matrix_coefficients>BT.601</matrix_coefficients>
``

dericed:
Just to verify if it's mediainfolib or the file, can you paste the output of the output of `ffmpeg -i file.mkv` and `ffmpeg -i file.mov`, particularly the line that starts with `Stream #0:0`

beccabender:
Matroska:

``Stream #0:0: Video: ffv1 (FFV1 / 0x31564646), yuv422p10le(tv, bt470bg/bt470bg/bt709, top coded first (swapped)), 720x576, SAR 16:15 DAR 4:3, 25 fps, 25 tbr, 1k tbn, 1k tbc (default)``

Quicktime: 

``    Stream #0:0(eng): Video: ffv1 (FFV1 / 0x31564646), yuv422p10le(smpte170m/bt470bg/bt709, top coded first (swapped)), 720x576, 1865 kb/s, SAR 16:15 DAR 4:3, 25 fps, 25 tbr, 12800 tbn, 12800 tbc (default)
``

beccabender:
Reports from ffprobe:

Matroska:
``
color_range=tv
 color_space=bt470bg  
color_transfer=bt709  
color_primaries=bt470bg  
chroma_location=unspecified
``

Quicktime: 
``
color_range=unknown
color_space=smpte170m
color_transfer=bt709
color_primaries=bt470bg
chroma_location=unspecified
``

dericed:
IMHO this issue can be closed. The difference in matrix_coefficients between QuickTime and Matroska is justified, since Matroska is able to be more specific than QuickTime can support. Both are using the more accurate matrix_coefficients terminology possible.

retokromer:
If no opposition in the next 24 hours, I’ll close this one.

kieranjol:
Perhaps not,seeing as this patch in ffmpeg could rectify the issue?
https://patchwork.ffmpeg.org/patch/6782/

On 15 Dec 2017 09:37, "Reto Kromer" <notifications@github.com> wrote:

[...]

PR #194, opened 2017-11-01
[WIP] Draft vrecord policy for FFV1 Matroska and Quicktime

beccabender:
This is a work-in-progress policy to validate FFV1 Matroska and Quicktime Files.

### DIFF
--- a/vrecord_policy.xml
+++ b/vrecord_policy.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
-<policy type="and" name="Check validity of makelossless files">
-    <description>This policy tests lossless FFV1 Matroska and Quicktime files made using vrecord and checks their validity</description>
+<policy type="and" name="Check validity of FFV1 Matroska and Quicktime files">
+    <description>This policy tests FFV1 Matroska and Quicktime files made using vrecord and checks their validity</description>
     <policy type="or" name="Is the container Matroska or Quicktime?">
         <policy type="and" name="Is it good Matroska?">
             <policy type="and" name="Is it new Matroska with an UUID?">
--- a/vrecord_policy.xml
+++ b/vrecord_policy.xml
@@ -49,7 +49,7 @@
             <rule name="Is the color space YUV?" value="ColorSpace" tracktype="Video" operator="=">YUV</rule>
             <rule name="Is the chroma subsampling 4:2:2?" value="ChromaSubsampling" tracktype="Video" operator="=">4:2:2</rule>
             <rule name="Is the video interlaced?" value="ScanType" tracktype="Video" operator="=">Interlaced</rule>
-            <rule name="Is the comression mode lossless?" value="Compression_Mode" tracktype="Video" operator="=">Lossless</rule>
+            <rule name="Is the compression mode lossless?" value="Compression_Mode" tracktype="Video" operator="=">Lossless</rule>
             <rule name="Is max slice count at least 4?" value="extra/MaxSlicesCount" tracktype="Video" operator="&gt;=">4</rule>
             <rule name="Is error detection type per slice?" value="extra/ErrorDetectionType" tracktype="Video" operator="=">Per slice</rule>
             <rule name="Is the video stream first?" value="StreamOrder" tracktype="Video" operator="=">0</rule>
--- a/vrecord_policy.xml
+++ b/vrecord_policy.xml
@@ -1,0 +1,93 @@
+<?xml version="1.0"?>
+<policy type="and" name="Check validity of makelossless files">
+    <description>This policy tests lossless FFV1 Matroska and Quicktime files made using vrecord and checks their validity</description>
+    <policy type="or" name="Is the container Matroska or Quicktime?">
+        <policy type="and" name="Is it good Matroska?">
+            <policy type="and" name="Is it new Matroska with an UUID?">
+                <rule name="Is it Matroska?" value="Format" tracktype="General" operator="=">Matroska</rule>
+                <rule name="Matroska version 4 or greater?" value="Format_Version" tracktype="General" operator="&gt;=">4</rule>
+                <rule name="Unique ID is present?" value="UniqueID" tracktype="General"/>
+                <rule name="Is the FFV1 Codec ID V_MS/VFW/FOURCC / FFV1?" value="CodecID" tracktype="Video" operator="=">V_MS/VFW/FOURCC / FFV1</rule>
+                <rule name="Is the PCM Codec ID A_PCM/INT/LIT" value="CodecID" tracktype="Audio" operator="=">A_PCM/INT/LIT</rule>
+                <!-- add Endianness and Signedness when added to report -->
+            </policy>
+            <policy type="or" name="Is it NTSC SD or PAL SD?">
+                <description>Quick test for NTSC vs Pal to validate matrix coefficients. Full NTSC vs PAL test below</description>
+                <policy type="and" name="Is it NTSC?">
+                    <rule name="NTSC FrameRate 29.970?" value="FrameRate" tracktype="Video" operator="=">29.970</rule>
+                </policy>
+                <policy type="and" name="Is it PAL?">
+                    <description>The PAL matrix coefficients value is reported differently depending on container. This test validates for Matroska</description>
+                    <rule name="PAL FrameRate 25.000?" value="FrameRate" tracktype="Video" operator="=">25.000</rule>
+                    <rule name="Are the matrix coefficients BT.470 System B, BT.470 System G?" value="matrix_coefficients" tracktype="Video" operator="=">BT.470 System B, BT.470 System G</rule>
+                </policy>
+            </policy>
+            <policy type="and" name="Is it good Quicktime?">
+                <rule name="Is the file extension mov?" value="FileExtension" tracktype="General" operator="=">mov</rule>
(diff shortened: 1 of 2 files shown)


privatezero:
LGTM!

kieranjol:
This looks really good and super useful.is the idea that this will auto
validate at the end of capture?

beccabender:
Yes, the goal is to have it evaluate the file following capture.

vrecord_policy.xml
```diff
@@ -0,0 +1,93 @@
...
+            </policy>
+            <rule name="Is the video format FFV1?" value="Format" tracktype="Video" operator="=">FFV1</rule>
+            <rule name="FFV1 is version 3.4 or later?" value="Format_Version" tracktype="Video" operator="&gt;=">3.4</rule>
+            <rule name="FFV1 is encoded in GOP size of 1?" value="Format_Settings_GOP" tracktype="Video" operator="=">N=1</rule>
+            <rule name="Is the display aspect ratio 1.333?" value="DisplayAspectRatio" tracktype="Video" operator="=">1.333</rule>
+            <rule name="Is the frame rate mode constant?" value="FrameRate_Mode" tracktype="Video" operator="=">CFR</rule>
+            <rule name="Is the color space YUV?" value="ColorSpace" tracktype="Video" operator="=">YUV</rule>
+            <rule name="Is the chroma subsampling 4:2:2?" value="ChromaSubsampling" tracktype="Video" operator="=">4:2:2</rule>
+            <rule name="Is the video interlaced?" value="ScanType" tracktype="Video" operator="=">Interlaced</rule>
+            <rule name="Is the comression mode lossless?" value="Compression_Mode" tracktype="Video" operator="=">Lossless</rule>
```
kieranjol: typo: comression

vrecord_policy.xml
```diff
@@ -0,0 +1,93 @@
...
+            <policy type="and" name="Is it good Quicktime?">
+                <rule name="Is the file extension mov?" value="FileExtension" tracktype="General" operator="=">mov</rule>
+                <rule name="Is the format MPEG-4?" value="Format" tracktype="General" operator="=">MPEG-4</rule>
+                <rule name="Is the format profile QuickTime" value="Format_Profile" tracktype="General" operator="=">QuickTime</rule>
+                <rule name="Is the FFV1 Codec ID FFV1?" value="CodecID" tracktype="Video" operator="=">FFV1</rule>
+                <rule name="Is the PCM Codec ID in24?" value="CodecID" tracktype="Audio" operator="=">in24</rule>
+                <rule name="Is the endianness Little?" value="Format_Settings_Endianness" tracktype="Audio" operator="=">Little</rule>
+                <rule name="Is the signedness signed?" value="Format_Settings_Sign" tracktype="Audio" operator="=">Signed</rule>
+            </policy>
+            <policy type="or" name="Is it NTSC SD or PAL SD?">
```
kieranjol: There is another check of this type up in line 14 that looks the same aside from the rec601 check, but perhaps this is intentional? I"m not really familiar with the and/or policies very well.

vrecord_policy.xml
```diff
@@ -0,0 +1,93 @@
...
+            <rule name="Is max slice count at least 4?" value="extra/MaxSlicesCount" tracktype="Video" operator="&gt;=">4</rule>
+            <rule name="Is error detection type per slice?" value="extra/ErrorDetectionType" tracktype="Video" operator="=">Per slice</rule>
+            <rule name="Is the video stream first?" value="StreamOrder" tracktype="Video" operator="=">0</rule>
+            <rule name="Is the colour description present?" value="colour_description_present" tracktype="Video" operator="=">Yes</rule>
+            <rule name="Are the transfer characteristics BT.709?" value="transfer_characteristics" tracktype="Video" operator="=">BT.709</rule>
+            <policy type="or" name="Is the bit depth 8 or 10?">
+                <rule name="Is the bit depth 8?" value="BitDepth" tracktype="Video" operator="=">8</rule>
+                <rule name="Is the bit depth 10?" value="BitDepth" tracktype="Video" operator="=">10</rule>
+            </policy>
+            <policy type="or" name="Is it NTSC SD or PAL SD?">
```
kieranjol: I wonder if the previous NTSC or PAL checks could all be moved in here? Looks like there is some duplication.

vrecord_policy.xml
```diff
@@ -0,0 +1,93 @@
...
+                <policy type="and" name="Is it NTSC?">
+                    <rule name="NTSC height 486?" value="Height" tracktype="Video" operator="=">486</rule>
+                    <rule name="NTSC Width 720?" value="Width" tracktype="Video" operator="=">720</rule>
+                    <rule name="NTSC FrameRate 29.970?" value="FrameRate" tracktype="Video" operator="=">29.970</rule>
+                    <rule name="NTSC BitRate Mode CBR?" value="BitRate_Mode" tracktype="Video" ooperator="=">CBR</rule>
```
kieranjol: Is this intentional? I see that the PAL check is looking for VBR.

vrecord_policy.xml
```diff
@@ -0,0 +1,93 @@
...
+                    <rule name="NTSC pixel aspect ratio 0.900?" value="PixelAspectRatio" tracktype="Video" operator="=">0.900</rule>
+                    <rule name="NTSC scan order bottom field first?" value="ScanOrder" tracktype="Video" operator="=">BFF</rule>
+                    <rule name="NTSC colour primaries BT.601 NTSC?" value="colour_primaries" tracktype="Video" operator="=">BT.601 NTSC</rule>
+                    <rule name="Are the matrix coefficients BT.601?" value="matrix_coefficients" tracktype="Video" operator="=">BT.601</rule>
+                </policy>
+                <policy type="and" name="Is it PAL?">
+                    <rule name="PAL height 576?" value="Height" tracktype="Video" operator="=">576</rule>
+                    <rule name="PAL Width 720?" value="Width" tracktype="Video" operator="=">720</rule>
+                    <rule name="PAL FrameRate 25.000?" value="FrameRate" tracktype="Video" operator="=">25.000</rule>
+                    <rule name="PAL BitRate Mode VBR?" value="BitRate_Mode" tracktype="Video" operator="=">VBR</rule>
```
kieranjol: As with line 67 - is this intentional, should both be looking for VBR?

dericed: I guess the uncompressed should be CBR, but for the others maybe the test is not needed. For instance if someone made an ffv1 recording of black, the result may intentionally be CBR and not VBR since all the frames would be identical.

vrecord_policy.xml
beccabender: I wasn't quite sure  the best way to go about this.... because the matrix coefficients values are different for PAL depending on whether the format is quicktime or matroska, there needs to be redundant testing somewhere in the policy. So the reason the same test exists at line 14 as 34 is because one is for matroska and the other quicktime. 

I'm happy to move that testing down into the section starting at line 62, but then I would need to add in duplicate tests there for matroska and quicktime to deal with the differing PAL matrix coefficients. But maybe that's more efficient? 

I'm new to this!

kieranjol: Ah, this makes sense now - I remember seeing this in your issue https://github.com/amiaopensource/vrecord/issues/193 - I can see why you had to work around it like this.. I have nothing else to add as I'm quite unfamiliar with the and/or policies so I'm not sure how best to proceed, i think what you did looks good to me , though.

vrecord_policy.xml
beccabender: @dericed, to confirm, you mean to remove lines 67 and 77 for this ffv1 policy, but to include CBR when the policy gets expanded to handle uncompressed files made in vrecord?

dericed: yep

vrecord_policy.xml
```diff
@@ -0,0 +1,91 @@
...
+                    <rule name="PAL pixel aspect ratio 1.067?" value="PixelAspectRatio" tracktype="Video" operator="=">1.067</rule>
+                    <rule name="PAL scan order top field first?" value="ScanOrder" tracktype="Video" operator="=">TFF</rule>
+                    <rule name="PAL colour primaries BT.601 PAL?" value="colour_primaries" tracktype="Video" operator="=">BT.601 PAL</rule>
+                </policy>
+            </policy>
+        </policy>
+        <rule name="Is the audio stream on track 2?" value="StreamOrder" tracktype="Audio" operator="=">1</rule>
+        <rule name="Is the audio PCM?" value="Format" tracktype="Audio" operator="=">PCM</rule>
+        <rule name="Is the bit Depth 24?" value="BitDepth" tracktype="Audio" operator="=">24</rule>
+        <rule name="Is the sample rate 48khz?" value="SamplingRate" tracktype="Audio" operator="=">48000</rule>
```
retokromer: Might this be `kHz`?

beccabender: Yes - thanks for the catch!

retokromer:
Can this be merged or is there more to come?

dericed:
I suggest not merging as there's not actionable in it yet. vrecord would have to add a condition to say if ffv1 with mov or mkv, then apply this after capture. Also would need to update to add the mediaconch dependency.

retokromer:
Makes sense. Thank you!

retokromer:
I suggest to merge this as a placeholder, even if this is not yet used in vrecord.

dericed:
closing in favor of https://github.com/amiaopensource/vrecord/pull/241

PR #197, opened 2017-12-05
ensure standards test only gets one result

dericed:
otherwise an update to Desktop Video breaks this. thx to Lorena and Pamela for finding this

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -283,8 +283,8 @@ BRNG
 # about the card being used. We can add to this function in the future to allow
 # the user to choose different video configurations.
 get_standard_values(){
-    PAL_value="$(bmdcapture -h 2>&1 | grep "PAL" | cut -d: -f 1 | sed 's/ //g')"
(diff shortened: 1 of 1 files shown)


dericed:
The grep failed before since the output of `bmdcapture -h` used to be:

```
Supported video input display modes and pixel formats:
         0:   NTSC                 	 720 x 486 	   29.97 FPS
         1:   NTSC 23.98           	 720 x 486 	  23.976 FPS
         2:   PAL                  	 720 x 576 	      25 FPS
```

but a blackmagic update changes that to

```
Supported video input display modes and pixel formats:
         0:   525i59.94 NTSC       	 720 x 486 	   29.97 FPS
         1:   525p23.98 NTSC       	 720 x 486 	  23.976 FPS
         2:   625i50 PAL           	 720 x 576 	      25 FPS
```

so the grep is updated to work for both.

dericed:
Ya it's insurance, since it was missing it before. With the Desktop video update the variable was set to `0 1` and ... caused some fails.

Issue #207, opened 2017-12-15
do not let record time to be to 0


GregH18:
I think this might already be an issue, but I'm not completely sure. We were running some tests through vrecord and QCTools reports were not being created. An error message appeared saying `No such file or directory` (see image below). The ffmpeg and ffplay reports/logs were created and the video was captured. I'm not sure if this was partially due to us setting the `set record time in minutes` to `0`.

![image03](https://user-images.githubusercontent.com/27779274/33963821-60ab5ee8-e024-11e7-9cfc-c169a9c08f08.png)

Thanks for all your help!

dericed:
Are you on the latest vrecord (released today)? Do you have `qcli` installed?

dericed:
`grinnie`!!!!

GregH18:
Let me double check this! 

Grinnie! lol @WinnieSchLin

GregH18:
This issue is only occurring when the `set record time in minutes` is `0`. 

Does this meant the issue is closed? Sorry for wasting time. New to this!

dericed:
I updated the issue name. Unfortunately vrecord does not yet support digitizing analog tape to files that are 0 seconds long.

GregH18:
That sounds good to me! Maybe we can call this the `grinnie error`?

dericed:
please test https://github.com/amiaopensource/vrecord/pull/210

PR #210, opened 2017-12-15
more duration tests

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/207#issuecomment-352114826


GregH18:
Tested and the error message looks great. Thanks!

PR #211, opened 2017-12-15
update hardware

bturkus:
duo 2 works with ffmpeg decklink!


retokromer:
Good to know, that it works now! Thank you, @bturkus!

retokromer:
Does this resolve https://github.com/amiaopensource/vrecord/issues/148?

Issue #217, opened 2017-12-19
Record mode broken if qctoolsxml output turned off


privatezero:
@libbyhopfauf discovered that if qctoolsxml output is turned off, record mode will just start piping into the terminal.

It looks like at https://github.com/amiaopensource/vrecord/blob/master/vrecord#L1209 the command was left unfinished and so vrecord just starts to cat the pipe all over the place!

retokromer:
I guess, it’s https://github.com/amiaopensource/vrecord/commit/f46a063226268764ccf418687e2d14ec40da1693

PR #220, opened 2018-01-03
fix bug when qctoolsxml output is turned off

retokromer:
trying to fix https://github.com/amiaopensource/vrecord/issues/217

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1205,8 +1205,8 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
         exit 1
     fi
     QCTOOLS_POST=(qcli -i - -o "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz")
-#else
-#    QCTOOLS_POST=(cat)
+else
+    QCTOOLS_POST=":"
 fi
     "${FFMPEG_DECKLINK[@]}" \
     2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
--- a/vrecord
+++ b/vrecord
@@ -1206,7 +1206,7 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
     fi
     QCTOOLS_POST=(qcli -i - -o "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz")
 else
-    QCTOOLS_POST=(cat >/dev/null)
+    QCTOOLS_POST=('cat >/dev/null')
 fi
     "${FFMPEG_DECKLINK[@]}" \
     2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
--- a/vrecord
+++ b/vrecord
@@ -1205,8 +1205,8 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
         exit 1
     fi
     QCTOOLS_POST=(qcli -i - -o "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz")
-else
-    QCTOOLS_POST=(cat)
+#else
+#    QCTOOLS_POST=(cat)
 fi
     "${FFMPEG_DECKLINK[@]}" \
     2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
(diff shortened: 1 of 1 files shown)


retokromer:
@libbyhopfauf Could you possibly test this commit? Thank you very much! (Currently I don’t have access to a setting.)

retokromer:
`>/dev/null`?

retokromer:
What about this one, @dericed?

privatezero:
Would it be more efficient to assign the `|` to a variable that could then create the pipe, or just end the command depending on the settings? That way it would avoid an extraneous process?

dericed:
I think I tested `>/dev/null` and it didn't work when placed as a variable. Same issue with `|` which is why I was using `cat` in that role, but I'm not sure why I missed what Libby found, perhaps I had the deck off during that test.

retokromer:
I’m quite sure (untested!), that a pipe to `:` would work.

retokromer:
In
```
#!/usr/bin/env bash

if [[ -n "${2}" ]]; then
  TEST=(cat)
else
  TEST=(:)
fi

echo "${1}" | "${TEST[@]}"
```
the `:` works as expected. Therefore I guess this would fix the issue.

dericed:
that looks good to me, i should have a chance to test locally soon

privatezero:
Sorry if I am totally missing something, but I tried running `ffmpeg -f lavfi -i testsrc -f nut - | tee >(ffplay -) | :` (which should approximate the chain in vrecord) and it seems that piping to `:` is killing the ffmpeg command and breaking the pipe to FFplay

dericed:
same here. i suppose another way to fix is to have two commands in a conditional if about if qctools is needed or not. I was hoping to avoid that redundancy but maybe it's not feasible or cleaner in another way.

retokromer:
Thank you very much @privatezero and @dericed for testing! Closing then my failed attempt.

retokromer:
This one works with @privatezero’s example.

retokromer:
False alarm, don’t work either. I guess we have to split into two commands.

retokromer:
FYI: I made some additional testings. `tee`and `:` or `>/dev/null` seem not to love each other.

privatezero:
For what it is worth I was able to get
`TEST=('cat >/dev/null')`
`ffmpeg -f lavfi -i testsrc -f nut - | tee >(ffplay -) | $(eval ${TEST[@]})`

to work, but I have no way of testing right now if adding the `eval` to the qctools output command would mess it up!

retokromer:
Thank you, @privatezero! I’ll give it another try.

retokromer:
Caveat: Please remember that I currently cannot test with a card!

This seems to work fine. On the `qcli` branch, I got the message:
```
qcli dev build, Oct 15 2017 20:29:44, Copyright (c) 2013-2017 BAVC  
Usage: qcli -i <qctools-input> [-o <qctools-output>]

Use qcli -h to get detailed help
```
in the terminal, which could probably also be killed.

retokromer:
I squash and merge this now.

@libbyhopfauf Could you possibly test the HEAD? Thank you?

If it solves https://github.com/amiaopensource/vrecord/issues/217, then I will prepare a new release.

dericed:
who tested this?

dericed:
I propose to revert this with https://github.com/amiaopensource/vrecord/pull/223. This pr had no approving review and only changes the nature of the error to terminate with `./vrecord: line 1227: cat >/dev/null: No such file or directory`

retokromer:
Indeed, but it solves one of the three bugs that are open since quite longtime.

dericed:
@retokromer are you sure this PR fixes https://github.com/amiaopensource/vrecord/issues/217? Is there any successful test?

retokromer:
I cannot test a regular setting now (nor in the next days), but with my laptop, an Intensity USB and an S-video PAL camera it works. Therefore I would like that Libby (or someone else) could test easily on her regular setting.

It seems to me that the best release we have is currently v2017-10-17, and I would like to have a more recent one, with your improvements, but without the bugs. I feel uncomfortable in recommending the two newer releases, therefore I was trying to work on improve it.

dericed:
I still propose reverting this merge via https://github.com/amiaopensource/vrecord/pull/223. The single-quoted redirection didn't work for me and no one approved this PR. I'll wait to check consensus but recommend reverting this.

retokromer:
I am very fine with reverting. This is not my point.

PR #221, opened 2018-01-03
fix alignment to main level

retokromer:


privatezero:
Should then https://github.com/amiaopensource/vrecord/blob/799f566a6fb6998054e9a1c65921185171f0dc5a/vrecord#L1227 also be realigned?

retokromer:
Indeed! Thank you, @privatezero! I will add.

retokromer:
No, this is actually the end of the command.

PR #223, opened 2018-01-08
Revert "fix bug when qctoolsxml output is turned off"

dericed:
Reverts amiaopensource/vrecord#220


dericed:
This is to revert #220. I also set the master branch to require reviews before merging.

PR #224, opened 2018-01-08
conditional ffmpeg rather than conditional pipe

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/217

workaround til a better route is found

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1204,27 +1204,41 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
         _report -w "Such as \`brew install qcli\`."
         exit 1
     fi
-    QCTOOLS_POST=(qcli -i - -o "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz")
+    "${FFMPEG_DECKLINK[@]}" \
+        2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
+        -v info -nostdin -hide_banner -nostats "${inputoptions[@]}" "${GRAB_DECKLINK[@]}" \
+        -metadata:s:v:0 encoder="${codecname}" "${middleoptions[@]}" -c:a pcm_s24le \
+        -filter_complex "\
+            [0:v:0]${recordingfilter};\
+            ${audiomap}" \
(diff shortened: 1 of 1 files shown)


privatezero:
I just tried testing

`ffmpeg -f lavfi -i testsrc -f nut - | tee >(ffplay -) | $(eval ${TEST[@]})`

After setting the TEST array to both
`TEST=('cat >/dev/null')`
and
`TEST=(qcli -i - -o "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz")`

and adding the the `eval` seems to make both commands work, but I can't test this within an actual vrecord test capture.

dericed:
`eval` can add security issues, but I'm glad with this way at least eval is restricted to a small part of the chain rather than the whole chain.

privatezero:
Hahaha as in the potential for something horrible to happen if someone used `rm -r $HOME` as their mediaid?

Anyway - down with making the FFmpeg command conditional also! Just thought I would throw that out there in case it seemed like a last viable train of exploration.

dericed:
@privatezero, exactly. Perhaps we could ensure that the mediaid adheres to some regex to avoid MEDIAID=" ; rm -r $HOME/andrews_mp3_collection". But I suggest to merge to fix the issue first and maybe we can have a better pr later.

vrecord
```diff
@@ -1204,27 +1204,41 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
...
-    $(if [ -n "$map1k" ] ; then echo ${map1k} ; fi) $(if [ -n "$map1v" ] ; then echo ${map1v} ; fi) \
-    $(if [ -n "$map2k" ] ; then echo ${map2k} ; fi) $(if [ -n "$map2v" ] ; then echo ${map2v} ; fi) \
-    -f "${format}" "${dir}/${id}${suffix}.${extension}" \
-    "${extraoutputs[@]}" -map 0 -c copy -f nut - | \
-    tee >("${FFPLAY_DECKLINK[@]}" \
-        -v info -hide_banner -stats -autoexit -i - \
-        -window_title "mode:${runtype} - video:'${video_input_choice}' audio:'${audio_input_choice}'" \
-        -vf "${playbackfilter}") | \
-        "${QCTOOLS_POST[@]}"
+
```
retokromer: I suggest only one empty line, not three of them.

Issue #208, opened 2017-12-15
"Invalid data" error: MXF/JPEG2000 and uncompressed


GregH18:
One more issue. We captured a video using an MXF container and JPEG2000 video codec, but we receive the "Invalid data" error message below. We're also having trouble with MXF/Uncompressed also. Can anyone confirm these problems?

![image04](https://user-images.githubusercontent.com/27779274/33963857-7ff8b782-e024-11e7-8b93-bcb45e5eed94.png)

retokromer:
Did anyone reproduce this?

retokromer:
@GregH18 Which version of vrecord are you running? (The command `vrecord -h` prints the installed version.)

bturkus:
hi @retokromer. Sorry for not getting back you sooner about this, but I'm seeing the error on vrecord 2018-01-05. 

When I run ```vrecord -h```, I get the following:
```
'vrecord 2017-06-15 ( /vrecord/2017-07-15 ( /vrecord/2017-07-27 ( /vrecord/2017-10-17 ( /vrecord/2017-12-15 ( /vrecord/2018-01-05
```
Do you know what all of this means, and is it a prob in any way?

retokromer:
@bturkus What is your output of `brew info vrecord`?

bturkus:
I was able to reproduce the error, for both jpeg200 and uncompressed. And here's what brew info says:

```
amiaopensource/amiaos/vrecord: stable 2018-01-05, HEAD
Capturing a video signal and turning it into a digital file
https://github.com/amiaopensource/vrecord
/usr/local/Cellar/vrecord/2017-06-15 (6 files, 136.9KB)
  Built from source on 2017-06-15 at 12:01:52
/usr/local/Cellar/vrecord/2017-07-15 (6 files, 137.0KB)
  Built from source on 2017-07-19 at 15:17:04
/usr/local/Cellar/vrecord/2017-07-27 (6 files, 137.3KB)
  Built from source on 2017-07-28 at 13:32:49
/usr/local/Cellar/vrecord/2017-10-17 (6 files, 135KB)
  Built from source on 2017-12-08 at 10:10:05
/usr/local/Cellar/vrecord/2017-12-15 (6 files, 134.8KB)
  Built from source on 2017-12-15 at 12:40:49
/usr/local/Cellar/vrecord/2018-01-05 (6 files, 135KB) *
  Built from source on 2018-01-08 at 13:36:09
[...]

PR #225, opened 2018-01-08
select the installed version only

retokromer:
fix the error reported by @bturkus in https://github.com/amiaopensource/vrecord/issues/208#issuecomment-356069101 i.e. all versions on the computer are listed and not only the installed one


bturkus:
thanks @retokromer!!!

retokromer:
Ha! I was hesitating as well, but then decided to not, because I don’t know Brew’s actual syntax.

Issue #227, opened 2018-01-10
Update documentation


retokromer:
- The changes made in the code should be reflected in the documentation.
- If one volunteers for `README.md`, I’ll volunteer for conforming `vrecord.1`.
- Once the documentation is updated as well, I suggest to release a new version .

PR #229, opened 2018-01-11
update documentation

retokromer:
Should resolve https://github.com/amiaopensource/vrecord/issues/227.

- update version
- update dependencies


retokromer:
Did I miss something? Otherwise I suggest to release a new version.

dericed:
The data of the patch seems true regarding of the releases.

PR #234, opened 2018-01-22
Adding option to switch between views

aeschweik:
Very WIP, adding the ability to switch between views in vrecord without exiting the program, as per the key in the top left corner below:

screen shot 2018-01-22 at 5 06 06 pm

The views are controlled by keys assigned in a Lua script (mpvs.lua, see "local scopes" for the assignments and "filters" for the code) that must be stored in the same directory as vrecord. If you run vrecord using mpv instead of ffplay (which is now an option you can choose in edit mode under "Select Media Player"), it will append the Lua script and you'll be able to switch between modes by pressing the assigned keys, all while vrecord does its thing.

Views that work: broadcast range visual, visual + numerical (don't ask about the Papyrus), color matrix, and the others listed in the screenshot. Views that I'm working on adding: full range visual and bit planes. Views that can be added: as many as there are keys on the keyboard!!

More examples: 6 color waveform
screen shot 2018-01-22 at 5 08 50 pm

7 overlaid waveform
screen shot 2018-01-22 at 5 08 52 pm

(The menu goes away eventually.)

The Lua script is almost 100% indebted to [this script](https://paste.ubuntu.com/25318437/) by Paul B Mahol.


privatezero:
sooooo dope

bturkus:
I was gonna just say "soooo dope." How dare you beat me to it!

bturkus:
Serious question: would it be possible to add a view that was just pure colorful fun? Like something [oioiiooixiii](https://www.youtube.com/watch?time_continue=2&v=Xgo07Xa2MsU)-styled? 

plz no patches welcome

vrecord
```diff
@@ -26,6 +26,7 @@ AUD_OUTLIER_THRSHLD=10
 BRNG_OUTLIER_THRSHLD=14
 FFMPEG_DECKLINK=($(brew --prefix ffmpegdecklink)/bin/ffmpeg-dl)
 FFPLAY_DECKLINK=($(brew --prefix ffmpegdecklink)/bin/ffplay-dl)
+SCRIPTDIR=$(dirname $(which "${0}"))
```
retokromer: I would suggest to move this before line 8, and to use the variable in the line 8 as well.

dericed:
@bturkus why not start the instagram fork of vrecord?

vrecord
retokromer: Where is `${SCRIPTDIR}` used?

aeschweik:
I believe that all the display options except for audiopassthrough are working! If anyone is willing to try out the scopes I would love to get feedback - select "Switch between views" in playback options.

I apologize for the messy commits and promise I will try again to fix later (rebasing is born of the devil).

privatezero:
I can't test within an actual vrecord session right now, but I tried out the different scopes just using the lua script and a sample file in mpv and they look great!

I realize that it still doesn't work, but I wonder if the audiopassthrough filter could be superseded by some audio specific views, like just a view for peak meters/audio phase! Since the lua script enables changing views on the fly, there might not be a need any more to have the meters superimposed on the standard view. (This might also make the audio monitoring less prone to causing lag in the system).

privatezero:
Or hell, maybe the [filter from audiorecorder](https://github.com/amiaopensource/audiorecorder/blob/master/audiorecorder#L388) could just be ripped straight out as a view in vrecord! (although it is pushing mpv to the limits even with a downsample to 44.1 kHz/16 bit audio). I had an older test filter chain set up for mpv instead of ffplay [over here](https://github.com/privatezero/NDSR/blob/master/live-audio-filter-test-scripts.txt#L77)

bturkus:
Not sure if it'd be helpful, but I've been using this recently and loving it:

https://github.com/mfcc64/mpv-scripts/blob/master/visualizer.lua

Lua scripts are awesome! Thanks for turning me on to them Annie. Can't wait to test out the new vrecord!

retokromer:
I am afraid, I cannot test this with a full setting at the moment. Yet what I can simulate/emulate LGTM. Thank you very much, @aeschweik! (I am a big fan of `mpv` and use it constantly in my company.)

aeschweik:
The commits are less messy now and hopefully should be less confusing for anyone who wants to try it out!

@privatezero I was looking at adding some other audio filter views but hadn't thought to look at audiorecorder, thank you for the reminder and the links!!

vrecord
```diff
@@ -1010,9 +1015,15 @@ else
 fi
 
 if [[ "${runtype}" = "passthrough" ]] ; then
-    "${FFPLAY_DECKLINK}" -v info -nostats "${GRAB_DECKLINK[@]}" 2> /tmp/vrecord_input.log \
+    if [[ "${media_player_choice}" = "ffplay" ]] ; then
```
dericed: I suggest using `if [mpv] xxx ; else yyy ; fi` rather than an if/elif with no else. In this case if the variable somehow does not equal ffplay or mpv then it will only exit. If using if/else then you could also skip `media_player_choice="ffplay"` as that would be covered by the `else`.

vrecord
```diff
@@ -202,7 +203,7 @@ video_codec_options=("Uncompressed Video" "FFV1 version 3" "JPEG2000" "ProRes" "
 channel_mapping_options=("2 Stereo Tracks (Channels 1 & 2 -> 1st Track Stereo, Channels 3 & 4 -> 2nd Track Stereo)" "1 Stereo Track (From Channels 1 & 2)" "1 Stereo Track (From Channels 3 & 4)" "Channel 1 -> 1st Track Mono, Channel 2 -> 2nd Track Mono" "Channel 2 -> 1st Track Mono, Channel 1 -> 2nd Track Mono" "quit")
 container_options=("QuickTime" "Matroska" "AVI" "MXF" "quit")
 standard_options=("NTSC" "PAL" "quit")
-playbackview_options=("Broadcast Range Visual" "Full Range Visual" "Visual + Numerical" "Color Matrix" "Bit Planes" "quit")
+playbackview_options=("Switch between views" "Broadcast Range Visual" "Full Range Visual" "Visual + Numerical" "Color Matrix" "Bit Planes" "quit")
```
dericed: Possibly could have a better name, perhaps `qcview (mpv)` with the lua script then renamed to `qcview.lua`. 

¯\\_(ツ)_/¯

vrecord
```diff
@@ -1184,67 +1194,74 @@ else
     middleoptions+=(-color_range mpeg)
 fi
 
-if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
-    if [[ ! "$(which qcli)" ]] ; then
```
dericed: This test was moved further down, but I suggest to keep it before the digitization commands. Otherwise the digitization commands will simply fail and provide a much messier output. Not blocking, but if unresolved here I'd suggest to resolve soon.

vrecord
```diff
@@ -1184,67 +1194,74 @@ else
...
+        SAT_OUTLIERS=$(gzcat "${QCXML}" | xml sel -t -v "count(//tag[@key='lavfi.signalstats.SATMAX'][@value>496])" -n)
     elif [[ "${pixel_format}" = "uyvy422" ]] ; then
-        SAT_OUTLIERS=$(gzcat "${logdir}/${id}${suffix}.${extension}.qctools.xml.gz" | xml sel -t -v "count(//tag[@key='lavfi.signalstats.SATMAX'][@value>124])" -n)
+        SAT_OUTLIERS=$(gzcat "${QCXML}" | xml sel -t -v "count(//tag[@key='lavfi.signalstats.SATMAX'][@value>124])" -n)
+    fi
+    AUD_OUTLIERS=$(gzcat "${QCXML}" | xml sel -t -v "count(//tag[@key='lavfi.astats.Overall.Max_level'][@value>=0.999998])" -n)
+    BRNG_OUTLIERS=$(gzcat "${QCXML}" | xml sel -t -v "count(//tag[@key='lavfi.signalstats.BRNG'][@value>=0.03])" -n)
+    audio_peak=$(gzcat "${QCXML}" | grep lavfi.astats.Overall.Peak_level | cut -d '"' -f 4 | sort -n | tail -n 1)
+    _writeingestlog "Peak Volume is (dB)" "${audio_peak}"
+    if [[ ! -s "${XML}" ]] ; then
```
dericed: Do you mean `${QCXML}`?

dericed: also the above lines that use QCXML could be included in this condition to only run if the QCXML file is there, rather than attempt to use QCXML several times and then show a warning if it's not present later.

mpvs.lua
```diff
@@ -0,0 +1,290 @@
+--[[
+mpv scopes
+
+Copyright 2017 Paul B Mahol
+License: public domain
+Source: https://paste.ubuntu.com/25318437/
```
dericed: This is changed a lot since @richardpl's post, so it may make sense to add a subsequent copyright notice for 2018 with the same license suggested.

vrecord
```diff
@@ -1184,67 +1194,74 @@ else
...
 
-if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
-    if [[ ! "$(which qcli)" ]] ; then
-        _report -w "Please install qcli to use the qctools reporting option."
-        _report -w "Such as \`brew install qcli\`."
-        exit 1
-    fi
+if [[ "${media_player_choice}" = "ffplay" ]] ; then
     "${FFMPEG_DECKLINK[@]}" \
-        2> >(tee "${logdir}/${id}${capturelogsuffix}" /tmp/vrecord_input.log >/dev/null) \
```
dericed: This chunk of changes is only based on indentation so I'd suggest to undo it to reduce changes.

dericed:
thx @aeschweik!

Issue #209, opened 2017-12-15
"Could not create DeckLink iterator" Error message


CSchloss385:
When I was installing the new vrecord  I came across the following error message: 
```
Could not create DeckLink iterator
dummy: Immediate exit requested
Please check connections and try again.
```
I was able to fix this problem by updating the Desktop Video through the Blackmagic Design website. Here's a link to the update: 

https://www.blackmagicdesign.com/support/family/capture-and-playback
I installed Desktop Video 10.9.7 

Just thought I would share since this might come up for other people as they download the new vrecord.

bturkus:
good call! does anyone know exactly what version (or higher) is required? I know it doesn't work with 10.9, but I haven't tried 10.9.3, 10.9.4, or 10.9.5.

dericed:
The requirements would vary based on your hardware.

iamdamosuzuki:
I'm getting this error running vrecord 2018-01-05 and using Black Magic software 10.6.7. This is the latest version of the BM software I can run with the Decklink 4K Extreme card. Later versions all cause input levels to clip at Broadcast Range, which according to Black Magic is a "known issue".

dericed:
Hi @iamdamosuzuki I have the latest Desktop Video with an Ultrastudio Express and it does not clip to broadcast range. Is this only a Decklink 4K Extreme issue? cc'ing @dheitmueller for any clues

iamdamosuzuki:
Howdy @dericed. I found the issue only happens with the analog inputs on the Decklink 4K Extreme. Our Ultrastudio Express doesn't have the problem.

retokromer:
Should the [needed hardware](https://github.com/amiaopensource/vrecord/blob/master/Resources/hardware.md) page be updated?

iamdamosuzuki:
@retokromer How would you suggest updating the page? Should we mention that any version after 10.6.X will not work with the Decklink cards? Is there any insight into what is causing the vrecord to not work with older versions of the Black Magic software? Until BM comes out with a patch for this (Which could be never), anyone using Decklink cards will be unable to use vrecord for analog transfers

retokromer:
>Should we mention that any version after 10.6.X will not work with the Decklink cards?

Yes, in my opinion, this would be useful.

>Is there any insight into what is causing the vrecord to not work with older versions of the Black Magic software?

I’m afraid, I don’t have any. And, sadly, I cannot delve into this now nor in the next days.

[...]

PR #236, opened 2018-01-27
add note regarding DeckLink 4K Extreme

retokromer:
context: https://github.com/amiaopensource/vrecord/issues/209#issuecomment-360862657


dericed:
could you add a citation here, even if its just hyperlinking the added text to a forum discussion. I think the addition could be seen as curious and it would be helpful to direct the reader to what inspired it.

retokromer:
It’s the discussion mentioned: https://github.com/amiaopensource/vrecord/issues/209#issuecomment-360862657

PR #248, opened 2018-02-07
start mono options

dericed:


dericed:
@privatezero, this commit was without a pr, any memory of the status?

dericed:
@privatezero ah it does cram the ui so the text overlaps, so needs a bit of x,y alignment

privatezero:
Hi! Yeah, sorry to slack on this - my mac has been at the office and I need to bring it back so I can test off the clock. I'll wrap this up this week so we can close! 

Related - do we really need options to create a mono track from either input? Do you think it would be sufficient to have a single option to create a mono track?

privatezero:
ok! I tweaked the gui layout a tad! I can't actually test the functions right now, but I ran a sample file with very distinct L/R tracks through that filter and can confirm that it is creating a mono track from the selected input channel.

privatezero:
@dericed just wanted to ping that this is probably ready for a merge if someone can test that it behaves as expected in an actual capture!

privatezero:
hahaha totally just accidentally requested a review from @aeschweik

aeschweik:
@privatezero 💁‍♀️ I would be happy to on my next day in the office!!

dericed:
thx @privatezero & @aeschweik

privatezero:
Yeah, i agree things are getting a bit cramped! I considered making changes to the gui, but that seemed more in scope with doing as a separate PR!

PR #250, opened 2018-02-08
little cleanup

retokromer:
Should improve readability of the code for maintenance:
- uniform alignment of playback filters
- uniform use of `elif` instead of `fi ; if` inside the same logical structure

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -308,7 +308,7 @@ duration_check(){
             _report -w "Illegal value for recording time. Input must only be numbers."
             exit 1
         fi
-        if [[ 1 -eq "$(echo "${duration} == 0" | bc)" ]] ; then
+        if [[ 1 -eq "$(echo "${duration} = 0" | bc)" ]] ; then
             _report -w "A recording duration of zero is invalid."
             exit 1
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -308,7 +308,7 @@ duration_check(){
             _report -w "Illegal value for recording time. Input must only be numbers."
             exit 1
         fi
-        if [[ 1 -eq "$(echo "${duration} == 0" | bc)" ]] ; then
+        if [[ 1 -eq "$(echo "${duration} = 0" | bc)" ]] ; then
```
dericed: duration="0.000" ; echo "${duration} = 0" | bc
(standard_in) 1: parse error

retokromer: Thank you, I’ll fix! (The behaviour is not clear to me from the top of the head.)

dericed: see https://www.gnu.org/software/bc/manual/html_mono/bc.html#SEC11

retokromer: Ah, I understand, it’s not the double [[. – Thank you very much, as usual!

PR #255, opened 2018-02-23
Adding a check for dropped frames

aeschweik:
Adding a built-in check for any dropped frames/signal interruptions during the vrecord process, pulling from decklink_input.log (where this information is recorded)!

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -486,6 +486,12 @@ yes_or_no(){
     esac
 }
 
+_frames_to_hhmmss(){
+    framerate=$(grep "fps=" "${logdir}/${id}${capturelogsuffix}" | cut -d' ' -f 5)
+    num=$(echo "scale=0; $i / $framerate" | bc)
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -486,6 +486,14 @@ yes_or_no(){
     esac
 }
 
+_frames_to_hhmmss(){
+    num=`expr $i / $framerate`
+    h=`expr "$num" / 3600`
+    m=`expr "$num"  % 3600 / 60`
+    s=`expr "$num" % 60`
```
retokromer: I suggest to use the `$( . . . )` syntax here as well, like at line 1317 and following.

vrecord
```diff
@@ -486,6 +486,14 @@ yes_or_no(){
     esac
 }
 
+_frames_to_hhmmss(){
+    num=`expr $i / $framerate`
```
dericed: expr 100 / 29.97
expr: not a decimal number: '29.97'

Issue #253, opened 2018-02-16
Add audio waveform after ingest


CSchloss385:
I think it would be helpful to have an audio waveform image generated after vrecord is finished running. This might be useful in better understanding where audio clipping is happening.

privatezero:
Were you thinking purely a sidecar file, or something more visual (like a window that opens automatically with the waveform, and maybe other QC data, after digitization)

CSchloss385:
Hmmm... I think it might be best to keep it as a sidecar file. Maybe it could be an option similar to the creation of framemd5 checksums where the user can decide if they want them generated or not.

retokromer:
+1 for @CSchloss385 suggestion

privatezero:
That way definitely seems cleanest! (Although I think with added options I definitely think we are inching towards needing a redesign of the GUI layout to reduce clutter)

CSchloss385:
That's true. the GUI is looking a little crowded!

privatezero:
Just tested out, and like I thought the `showwavespic` filter doesn't like to make an image from a capture (ran `rec -p | ffmpeg -i - -lavfi showwavespic=split_channels=1:s=1024x800 ~/Desktop/test.png` to test) so this would need to be an additional discrete event after the capture is complete.

Does adding a conditional to generate this post-record and pre-cowsay warnings make sense to people?

CSchloss385:
That makes sense to me!

privatezero:
Cool - I'll take a stab at this with using `drawtext` to add a some time context to the standard `showwavespic` output. Might make the results a little more usable!

dericed:
resolved in https://github.com/amiaopensource/vrecord/pull/357

PR #262, opened 2018-02-28
Generate audio waveform

privatezero:
Ok, I think this works to add in https://github.com/amiaopensource/vrecord/issues/253, but as per usual I can't test in an actual capture context!

Please somebody check it out and let me know if I have gone horribly wrong somewhere!


CSchloss385:
Thanks @privatezero! I'll try to test this today.

privatezero:
Cool! It was working sans GUI and with a fake input file on my linux machine, but I wasn't able to test it in combination with the GUI on my mac.

If it does work on your end, let me know if there are any changes (such as size) that would make it more useful. I restrained myself from jazzy colors as I didn't want to run afoul of @dericed. Although cyan is calling...

retokromer:
When possible, I always suggest to choose colours that works also for impaired persons.

privatezero:
In this case I just left it as FFmpeg defaults and figured changes could be made with any feedback. Are there any color combinations you know of that would be better for this @retokromer?

aeschweik:
Oops @jfarbowitz I don't know how to use GitHub - no review required!!

retokromer:
@privatezero I have to check first what FFmpeg’s standard is… More soon.

retokromer:
@privatezero We generally changed e.g. the colour for OK from green to blue in our interfaces. This way much more persons can distinguish it from the error’s red.

privatezero:
Ah - sorry for not being specific! FFmpeg creates a red and green waveforms (which I suppose might be the worst possible combination from an accessibility standpoint?)

retokromer:
Red and blue would be a much better choice!

privatezero:
@retokromer makes sense! I'll see if any other changes are required and then make a tweak for colour.

jfarbowitz:
No problem @aeschweik. It was nice to relive vrecord development, at least for a few minutes. 😄

CSchloss385:
Hi! Sorry this took me so long to test! I was able to get a waveform to show up, but I did get some errors related to the "sub-length component for drawtext" section. The error looks like this: 

usage: date [-jnu] [-d dst] [-r seconds] [-t west] [-v[+|-]val[ymwdHMS]] ... 
            [-f fmt date | [[[mm]dd]HH]MM[[cc]yy][.ss]] [+format]

As a result there are no timestamps on the waveform image. I think this might be a linux/mac compatibility issue? Not 100% sure though.

Also, I recommend switching the FreeSerif font to the default font since there is a default font already in vrecord. I was having a problem having it work with the FreeSerif font, but it worked fine with the default one.

privatezero:
Ok cool! Yeah - I totally meant to change that to the vrecord default font, but spaced out as usual! I know that `date` does have some annoying differences with mac and linux - I'll take a look at this when I get home! Thank you for testing!!

privatezero:
Hey @CSchloss385! Tweaked this so that it should work in macOS, but now there are a TON of upstream conflicts! For cleanness and simplicity, I will probably close this out and add the changes to a new PR from the updated head. I'll tag you again in that!

aeschweik:
Sorry about all the conflicts, @privatezero!! Just motivated to clean out all the PRs after the weekend.

privatezero:
hahaha @aeschweik, nah! Conflicts are good! Means that things are happening!! (And that I should have finished this up faster :-P)

dericed:
This branch is rebased to resolve conflicts. Please re-review.

privatezero:
@CSchloss385 @aeschweik, have you had a chance to see if this is all hunky dory post rebase?

CSchloss385:
I just ran a couple of tests and after the dropped frames analysis, I get the following message: 

grep: /: Is a directory
(standard_in) 2: parse error
grep: /: Is a directory
(standard_in) 2: parse error

This message occurs after the message "Warning: FFmpeg reported missing frames. The file may have sync issues."
It looks like it might be having a problem parsing the QC Tools xml? 

Everything else was working fine!

privatezero:
Were there any logs generated? Maybe the path to the logs it is trying to parse in `"${LOGDIR}/${ID}${CAPTURELOGSUFFIX}"` is somehow being set to `/` (maybe the settings got scrambled in the test environment?)

https://github.com/amiaopensource/vrecord/blob/c348f42d9ff242646128cc6fb9a656b89a8e9395/vrecord#L1357

CSchloss385:
Hi @privatezero ! Sorry this took so long to get back to you. Logs did generate, but I actually think the problem is related to a bug in the dropped frames reporting.

privatezero:
No worries @CSchloss385!  Thanks for testing! I suppose if the problem isn't related to the wave form generation, it is up to y'all to decide if it is good to merge or if @dericed wants to rebase it further!

dericed:
Can we reuse the new frames2hhmmss function from https://github.com/amiaopensource/vrecord/blob/eb4e3877002b0eb1375165342772dfd7f6688077/vrecord#L453-L457 instead of merging a 2nd one within this pr.

privatezero:
Maybe? It doesn't look like apples to apples, but it is probably doable.

I guess it would involve finding the total number of frames (instead of the total number of seconds as is currently used) and then using those to calculate the frame at the 1/4, 1/2 and 3/4 points. That frame could then be run through the `_frames_to_hhmmss` function to establish the pertinent time. Is the total number of frames also contained in the log?

The only other difference is that the filter chain for creating the waveform uses separate variables for `HH`:`MM`:`SS` instead of having them all combined as `HH:MM:SS` in order to avoid annoying problems of escaping the `:` in the variables contained in the filter chain. Either we could use a different method of escaping the colons, or the  `_frames_to_hhmmss` output could be parsed into separate variables as needed?

This would be nice as it would eliminate the extra `ffprobe` call, but as I can't actually live test right now someone else might have to take care of making and testing the changes.

privatezero:
Alternately, we could change the function to have a flag for frames _or_ seconds where it would crunch the math differently depending on if it was pointed at a number in frames or a number in seconds?

dericed:
Psst, @privatezero, I've been working on an script to make an image from an audio file to aid with qc. The draft of that script is at https://github.com/mediamicroservices/mm/blob/audiotest-drafts/makeaudiographs. I've been thinking of using that same script towards the objective of this pr as well. Here's a draft of the output, comments welcome.
hbr00097_01 wav_audio_report2

privatezero:
I think this looks lovely!! And it is certainly more informational than just using `showwavespic`. I am curious as to the logic for having an unprocessed and a processed visualization of sample diffs - is the hope that comparing them would help to isolate the real drops from the false positives? I am not sure I would explicitly label that section `interstitial` maybe `interstitial....???` ;-P

With the null tests, are you flipping the phase of a channel and summing to mono to visualize what would be lost should the original signal be summed?

Have you considered incorporating any frequency visualizations?

Love the colors - maybe needs a little cyan though? ;-P ;-P

retokromer:
Any news on this?

dericed:
Hi @retokromer afaict there is no news on this

privatezero:
@retokromer I think that this should be functional in its current form but I can't test on my end, so any final merge decisions would have to be by someone else. (And I see that there are conflicts again that would need resolution). However, it would definitely be superseded by @dericed's more comprehensive analysis should that be incorporated! I personally would vote for integrating that script in some form once it is done.

retokromer:
@privatezero I’m very fine either way, I just would like to have this available in `vrecord`… soon! @dericed’s approach sounds good to me too.

retokromer:
A gentle pinging on this.

dericed:
superseded by https://github.com/amiaopensource/vrecord/pull/357. If there's any missing gaps between this pr and that, please open an issue.

PR #263, opened 2018-02-28
pashua functionality: change radiobutton to popup

aeschweik:
Pashua tweaks in #248 made it impossible to select some buttons in other columns, so changed audio mapping options to a dropdown. (Sorry to PAL users for not catching this on the first go-round!)


privatezero:
Ah crap! Thanks for fixing that! The code looks good to me!

retokromer:
The code looks good to me as well.

PR #264, opened 2018-02-28
Embedding logs in Matroska files

aeschweik:
Adding embedded logs alongside the sidecar files for those working with Matroska!

This step currently takes place after the MediaConch policy evaluations because MediaConch parses out the attachments themselves to evaluate them as video files (a reported bug).

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -770,6 +770,15 @@ FRAMEMD5_CHOICE.default = ${FRAMEMD5_CHOICE}
 FRAMEMD5_CHOICE.option = ${UNDECLAREDOPTION}
 FRAMEMD5_CHOICE.option = Yes
 FRAMEMD5_CHOICE.option = No
+# embed logs choice
+EMBED_LOGS_CHOICE.x = 660
+EMBED_LOGS_CHOICE.y = 143
+EMBED_LOGS_CHOICE.type = radiobutton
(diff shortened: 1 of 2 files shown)


privatezero:
At the risk of cluttering the GUI that much more, maybe this should have an option to turn it off and on? 

That way people who are storing the logs as sidecars wouldn't get hit with double the storage space if they didn't want to? (I know as a percentage of the video file the logs are small, but some of them get pretty big!)

retokromer:
>maybe this should have an option to turn it off and on?

+1

kieranjol:
+1 on this being an option to turn off and on, though i could see myself turning it on when we finally commit to using vrecord. This is really cool work!

vrecord
```diff
@@ -1124,6 +1140,17 @@ else
     done
 fi
 
+if [[ "${embed_choice}" != "${undeclaredoption}" ]] ; then
+    echo ""
```
dericed: ?

aeschweik: I was copying the syntax of similar existing options (like framemd5_choice) and didn't stop to think on that line - probably unnecessary/worth changing in all of those?

dericed:
requires 'resolve conflict' please rebase

aeschweik:
Conflicts resolved!

PR #265, opened 2018-03-03
unify syntax

retokromer:
- use `[[ . . . ]]` instead of `[ . . . ]` when possible
- use `$( . . . )` instead of `` ` . . . ` ``
- delete not needed `;`


retokromer:
This has been included in https://github.com/amiaopensource/vrecord/pull/267.

PR #268, opened 2018-03-05
grep xml before xml sel

aeschweik:
removes a chunk of unnecessary lines before xmlstarlet runs, to reduce processing time (addressing #235)

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1264,12 +1264,12 @@ if [[ "${qctoolsxml_choice}" = "Yes" ]] ; then
     _report -d "Vrecord is analyzing your video file. Please be patient."
     if [[ -s "${QCXML}" ]] ; then
         if [[ "${pixel_format}" = "yuv422p10" ]] ; then
-            SAT_OUTLIERS=$(gzcat "${QCXML}" | xml sel -t -v "count(//tag[@key='lavfi.signalstats.SATMAX'][@value>496])" -n)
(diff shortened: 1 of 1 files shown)


CSchloss385:
Thank you @aeschweik !

dericed:
I was trying to neaten this by matching the full name of the key that you need to exclude all others, whereas the current draft just excludes based on a certain letter. regex has negative lookahead method to match like this, but the mac grep doesn't support it well.

However this works with perl:
`gzcat FILE.qctools.xml.gz | perl -nle 'print if not m{lavfi.(?!aphasemeter.phase)}' -`
Thus all `<tag>` elements that aren't `lavfi.aphasemeter.phase` get excluded, which really reduces the xml.

retokromer:
- Does `egrep` not give more possibilities on macOS as well? (Untested, just asking for the top of the head.)
- Indeed, the good old Perl works fine almost always!

dericed:
@retokromer, IIRC the perl process also worked faster than the grep process (before switching to use a negative lookahead). I did test with gnu grep (ggrep via `brew install grep`) but was unable to get it to work with negative lookahead.

retokromer:
@dericed Perl has (had?) the fastest regex implementation.

retokromer:
Thank you very much, @aeschweik! I’m confident that `xpath.c` will not longer fuse.

Issue #44, opened 2016-02-17
Modification to BRNG Pixels


bturkus:
I'm interested in having pixels valued at 0 and 255 turn a different color (red?) to indicate clipping. Or possibly some sort of gradient?

I know drop outs and edge related stuff would probably register, but we transfer a lot of 1/2'' open reel, which generally had no regard for legal limits. There are some cases where it seems better to take partial advantage of the allowances of head and footroom, though it can be tricky to know exactly when you've gone too far.

Dave suggested that files treated in such a way should be marked "Full Range." Could this be added as a preference?

dericed:
I think this is all doable with conditions in lutyuv. For now you could use the 'sample range' filter to select either 0 or 255. Whether a file is full range or broadcast range should be visible in the list view of qctools, in the column called "Color Range".

dericed:
please test https://github.com/amiaopensource/vrecord/tree/switch-from-brng-to-geq

kellyhaydon:
I tested today with a HI8mm color bars tape and exploded the values so they exceed 255 and 0, but there doesn't seem to be a difference in the visualizations with the exception of a slight color difference in the Waveform graphics. 

![geq](https://cloud.githubusercontent.com/assets/9436241/13270592/0aefe07e-da44-11e5-9bd4-3496d1ba17d2.png)
![brng](https://cloud.githubusercontent.com/assets/9436241/13270594/0c422694-da44-11e5-8f2d-6e00b2e5e6e1.png)

dericed:
Looks like it doesn't hit 0 and 255 but limits at 1-254.

dericed:
updated in https://github.com/amiaopensource/vrecord/commit/228906e16e00d172dc023fa9717e1e00bb9e94f4, retest plz

dericed:
also note that the geq equation only highlights luma, not chroma

retokromer:
What is the status of this?

richardpl:
I'm writing pseudocolor filter, which would, among other things, replace out-of-range pixels with different colored gradient. Perhaps this is what you need?

retokromer:
>I'm writing pseudocolor filter, which would, among other things, replace out-of-range pixels with different colored gradient.

Is one working on using the pseudocolor filter in vrecord?

aeschweik:
Closed via #269 thanks to @EG-tech!

PR #269, opened 2018-03-06
add pseudocolor to full range visual

EG-tech:
replaces lutyuv filter with pseudocolor filter for Full Range Visual to highlight full range pixels in red

settles issues #173 and #44 - makes pixels "out of range" (hitting 1 and 254 extremes) more noticeable

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -533,7 +533,7 @@ split=5[a][b][c][d][e];\
 [c1]${WAVEFORM_FILTER}[c2];\
 [a][b2][c2]vstack=inputs=3,format=yuv422p[abc1];\
 [d]${VECTORSCOPE_FILTER}[d1];\
-[e]format=yuv444p,lutyuv=y=if(gte(val\,254)\,1\,if(lte(val\,1)\,254\,val)):u=val:v=val,scale=512:ih[e1];\
+[e]format=yuv444p,"pseudocolor="'if(between(1,val,amax)+between(val,254,amax),65,-1):if(between(1,val,amax)+between(val,254,amax),100,-1):if(between(1,val,amax)+between(val,254,amax),212,-1)'"",scale=512:ih[e1];\
 [e1][d1]vstack[de1];\
 [abc1][de1]hstack"
           ;;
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -533,7 +533,7 @@ split=5[a][b][c][d][e];\
 [c1]${WAVEFORM_FILTER}[c2];\
 [a][b2][c2]vstack=inputs=3,format=yuv422p[abc1];\
 [d]${VECTORSCOPE_FILTER}[d1];\
-[e]format=yuv444p,lutyuv=y=if(gte(val\,254)\,1\,if(lte(val\,1)\,254\,val)):u=val:v=val,scale=512:ih[e1];\
+[e]format=yuv444p,"pseudocolor="'if(between(1,val,amax)+between(val,254,amax),65,-1):if(between(1,val,amax)+between(val,254,amax),100,-1):if(between(1,val,amax)+between(val,254,amax),212,-1)'"",scale=512:ih[e1];\
```
dericed: The double-quotes aren't needed here, you're already within a double-quote starting on line 528. Are you sure this works without escaping the commas?

EG-tech: I kept trying various combinations of quoting to escape the commas, and this *was* the only arrangement that kept visual from crashing on launch. Now suddenly it's not working again, so I will check back in tomorrow and try to see what's different from the line I had working earlier.

privatezero: FWIW, that chain works as part of a filter chain on my end with getting rid of all the double quotes and leaving the singles (which I think is the functional equivalent of `""'`...`'""`?)- so 

`pseudocolor='if(between(1,val,amax)+between(val,254,amax),65,-1):if(between(1,val,amax)+between(val,254,amax),100,-1):if(between(1,val,amax)+between(val,254,amax),212,-1)'`

also, killing all the quotes and escaping every comma with `\` seems to do the trick

EG-tech: confirmed, no quotes at all + escaping commas works. got thrown off since the example for pseudocolor in the ffmpeg documentation had double + single quotes, but that must have something to do with the gradient function that doesn't matter here? fixing this line, anyway!

dericed:
thx @EG-tech!

privatezero:
yay! Now @EG-tech needs to add `psuedocolor` to ffmprovisr!!

PR #278, opened 2018-03-12
add creation_time to files created using vrecord

aeschweik:
does not work for avi


aeschweik:
Oops, thought I had already squashed, thanks for the heads-up!

PR #280, opened 2018-03-14
check whether matroska has finished writing

aeschweik:


dericed:
minor, but i suggest consolidating the mkv tests into the `<policy type="and" name="Is it good Matroska?">` area

PR #290, opened 2018-03-26
Reversed NTSC and PAL sections

CSchloss385:
NTSC and PAL specs were accidentally switched in the lookup standard section.


dericed:
fixes a regression introduces in https://github.com/amiaopensource/vrecord/commit/94e9fa14b3e441fd3984ffa2bd1de70ccfdcda41

PR #299, opened 2018-04-11
just gui things

aeschweik:
Relaunches the GUI after edit and passthrough modes, rather than exiting. Selecting “record” and “exit” in the GUI works as usual.


EG-tech:
Hooray!!! I tested and all seems to work - "Passthrough", "Audio Check" and "Edit Settings" all return to the GUI when exiting.

I am getting an "image path is invalid" error in the GUI window for the vrecord_logo.png file, but maybe that's just because I'm not testing it in /usr/local and the prefix path is broken?

aeschweik:
Yeah, I'm pretty sure it's just breaking the `logo_path` - works fine when I run vrecord instead of ./vrecord, so should be resolved once it's run out of /usr/local!

PR #304, opened 2018-04-16
Reduce log verbosity

aeschweik:


dericed:
Unless there's a method (or hint that ffmpeg could resolve http://trac.ffmpeg.org/ticket/7141) I suggest we drop use of FFREPORT and instead generate the name of the output log as a variable and then simply use the `>2 (tee ...)` process as already included. This would lose the full ffmpeg command as FFREPORT reports it while the stderr doesn't, but perhaps we could log that somewhere else.

aeschweik:
Updated to do the above!

PR #306, opened 2018-04-20
Bug fix: Dropped frames

aeschweik:
Fixing a bug I introduced in my own bug fix (eb4e3877002b0eb1375165342772dfd7f6688077), cool


retokromer:
I guess, it’s what André Gide called _mise en abîme_ ;-)

PR #308, opened 2018-04-23
Dropped frames test: correct framerate

aeschweik:
Inconsistent padding in ffmpeg log breaks dropped frames test, plus the test was finding and using the encoding framerate rather than the standard framerate all along.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -586,6 +586,7 @@ _lookup_standard(){
     case "${1}" in
         "NTSC")
             STANDARD="ntsc"
+            DECKLINK_FPS="30000/1001"
             RECORDINGFILTER="setfield=bff,setsar=40/27,setdar=4/3"
             MIDDLEOPTIONS+=(-color_primaries smpte170m)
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -612,8 +614,7 @@ _yes_or_no(){
 }
 
 _frames_to_hhmmss(){
-    FRAMERATE=$(grep "fps=" "${LOGDIR}/${ID}${CAPTURELOGSUFFIX}" | cut -d' ' -f 5)
-    NUM=$(bc <<< "scale=0; ${i} / ${FRAMERATE}")
+    NUM=$(bc <<< "scale=0; ${i} / ${DECKLINK_FPS}")
```
dericed: maybe simpler as `NUM=$(bc <<< "${i} / (${DECKLINK_FPS})")` and then leave `DECKLINK_FPS="30000/1001"` and `DECKLINK_FPS="25000/1000"`

Issue #200, opened 2017-12-11
Error when starting vrecord: "Unsupported number of audio channels." 


iamdamosuzuki:
I did a goof and put this in the QCTools issue tracker. Here's what is known so far:

vrecord won't start up and ffplay returns the following error. 

`SDL_OpenAudio (8 channels, 4800 Hz): Unsupported number of audio channels.`

I can only make it happen if Blackmagic DeckLink 4K Extreme is selected as the Audio Output Device in the Apply System Preferences.

![33779085-552147a4-dc00-11e7-8913-469bd887026b](https://user-images.githubusercontent.com/3967767/33846000-1853ff5e-de5a-11e7-8828-e3358a33718e.png)

If the Audio Output is set to Built-In or Headphones the problem doesn't happen.

PR #309, opened 2018-04-23
set ffplay to only use the first two channels for playback

dericed:
SDL_OpenAudio does not support 8 channel audio

resolves https://github.com/amiaopensource/vrecord/issues/200


retokromer:
Can I merge this?
(I would like to make a new release and to update the formula.)

dericed:
any testing?

retokromer:
Don’t mind, I’ll fork.

dericed:
tested and merged

PR #315, opened 2018-05-03
force ProRes 422 HQ rather than default auto

retokromer:


dericed:
Any benchmarking on this? Is a realtime sd prores hq encoding more likely to fail?

retokromer:
- Hmm… I guess, on a slow computer the risk to fail is higher, on a fast computer the quality is certainly better.
- I had the idea to force, because this morning with experimental video from EPFL we got a… ProRes 4444 file!

retokromer:
The message is:
```
[prores_ks @ 0x7fe4a9831800] Autoselected 4:4:4:4 profile because of the used input colorspace. It can be overridden through -profile option.
```

dericed:
Is the 4444 autoselection from vrecord or did you encode a source with an alpha channel? Please share full uncut command and console output for such a case.

PR #317, opened 2018-05-04
Debugging help

aeschweik:
Not sure whether frame numbers match? Need to test further


retokromer:
It’s an excellent idea to add the version to the log!

vrecord
```diff
@@ -611,7 +611,7 @@ _yes_or_no(){
 
 _frames_to_hhmmss(){
     NUM=$(bc <<< "scale=0; ${i} / (${DECKLINK_FPS})")
-    printf "%02d:%02d:%02d\n" $(($NUM/3600)) $(($NUM%3600/60)) $(($NUM%60))
+    printf "%02d:%02d:%02d.%02d\n" $(($NUM/3600)) $(($NUM%3600/60)) $(($NUM%60)) $(($i%($DECKLINK_FPS)))
```
dericed: These numbers are off. I think it's probably not sane to keep it in integers. This is certainly messier but gets accurately from frame number to hh:mm:ss.mmm. With this the `NUM=` equation can be cut as well.

```
H=$(echo "$i / (60 * 60 * $DECKLINK_FPS)" | bc)
M=$(echo "($i - ($H*60 * 60 * $DECKLINK_FPS)) / (60 * $DECKLINK_FPS)" | bc)
S="$(echo "scale=3;($i - ($H*60 * 60 * $DECKLINK_FPS)-($M*60*$DECKLINK_FPS)) / ($DECKLINK_FPS)" | bc)"
Ss=$(echo "$S" | cut -d. -f1)
Sm=$(echo "$S" | cut -d. -f2)

printf "%02d:%02d:%02d.%03d\n" "$H" "$M" "$Ss" "$Sm"
```

aeschweik:
@dericed integrated your code!

PR #325, opened 2018-06-06
replace `vplay` -> `vtest`

retokromer:


retokromer:
I noted it while preparing today’s training… I didn’t see it during the PR, because I have been focussed on functionality.

PR #331, opened 2018-06-11
fix vtest version

dericed:
since there's no vtest formula, running vtest would give:

```
Warning: homebrew/core is shallow clone. To get complete history run:
  git -C "$(brew --repo homebrew/core)" fetch --unshallow

Error: No previously deleted formula found.
```


retokromer:
@dericed I just saw https://github.com/amiaopensource/vrecord/blob/8c29bc5005c36b64bb14a5fd575111fc90c9d138/vtest#L137 which I suggest you add to this PR.

Issue #333, opened 2018-06-12
review recommendation to make QCTools XML after capture


dericed:
Since https://github.com/amiaopensource/vrecord/releases/tag/v2018-06-11 and QCTools 0.9.2 the issues of running qcli and capturing at the same time seem to be resolved. @CSchloss385 and @savcampbell have tested this so far. I suggest we remove the recommendation.

retokromer:
+1

PR #335, opened 2018-06-13
Segment docs

EG-tech:
as discussed in #314


Resources/Documentation/troubleshooting.md
```diff
@@ -56,10 +56,6 @@ Error while decoding stream #0:0: Invalid data found when processing input
 
 You can safely ignore this warning, it's just FFmpeg complaining that it didn't receive a full frame of video when vrecord stopped.
 
-#### Sync Issues with FFmpeg 4.0 and qcli
```
dericed: There is a similar statement in settings.md, if removed, I suggest they be removed together.

Resources/Documentation/installation_and_setup.md
```diff
@@ -0,0 +1,88 @@
+## Installing vrecord
+
+If you use macOS, you can easily install vrecord using a package manager called Homebrew. To install Homebrew, follow [these instructions](https://brew.sh/).
```
retokromer: >If you use macOS

It‘s not needed, as the classic `vrecord` is running only under macOS.

dericed: i suggest this is not blocking, content changes could be in a subsequent pr

Resources/Documentation/troubleshooting.md
EG-tech: Should a line instead be included to explain why the option exists? i.e., it is recommended to create *during* capture but the option to create after can be used to troubleshoot. I can see a user being confused as to how/why they would pick one over the other.

EG-tech: Also, the warning in settings.md was not mine, so I was trying to just get back to structural revision, dealing with content subsequently as you suggested

dericed: Replacing the warning with an explanation sounds fine to me, whether in this pr or a subsequent one.

EG-tech: I'll just remove the settings.md warning for now for consistency and to resolve #333 , can move feature clarification/further content audit into a subsequent PR

dericed:
thx @EG-tech!!!!

PR #338, opened 2018-06-16
use C-style arithmetic evaluation

retokromer:


retokromer:
Oops, my bad! Thank you!

PR #344, opened 2018-06-21
add more audio input options

dericed:


dericed:
gonna close, since the additions aren't aligned well in the gui and the added options didn't actually fix the issue that inspired them (ie 'analog' worked as an input and the clarification for _rca vs _xlr wasn't needed)

PR #346, opened 2018-06-22
Mix of updates

dericed:
Attempt to resolve buffer overruns and timer issues. These commits add a `-v` option to increase the verbosity of the ffmpeg logging. Also the decklink timestamps now use the wallclock option for timing and the real time buffer is set to a larger value than the default. Ping to @libbyhopfauf @savcampbell @asfinn for testing.


savcampbell:
Tested this out today with the vrecord -v option and was still having issues with both the timer and buffer overruns.

retokromer:
@dericed: Should we possibly ask about the results of _Disk Speed Test_ as well?

dericed:
The scope of this PR is adjust to only add an option for verbose recording. Re-review needed.

dericed:
updated to resolve merge issue

Issue #349, opened 2018-06-28
embedded pal ffv1/mkv mediaconch fail


bturkus:
hi,
I'd love confirmation of this, but we're noticing that PAL ffv1/mkvs with embedded logs seem to be failing the built-in mediaconch test. We're not seeing this embedded ntscs, and I'm generally confused about why the files are failing. vrecord reports this:

![pal fail](https://user-images.githubusercontent.com/8987230/42051777-ef52e034-7ada-11e8-8528-a09f6e800f31.jpeg)
But the files themselves seem to be fine when tested in mediaconch proper:

![screen shot 2018-06-28 at 1 42 56 pm](https://user-images.githubusercontent.com/8987230/42051826-116ed420-7adb-11e8-8e53-839f1e5d95cf.png)

And this may be overkill, but here's the full xml produced by vrecord:
```
<?xml version="1.0" encoding="UTF-8"?>
<MediaConch xmlns="https://mediaarea.net/mediaconch" xmlns:mmt="https://mediaarea.net/micromediatrace" xmlns:mi="https://mediaarea.net/mediainfo" version="0.3">
  <media ref="/Volumes/NYPL_15905/2018_084_pami_090_wilson1/embed_test_pal_ffv1.mkv">
    <policy name="Check validity of FFV1 Matroska and Quicktime files" type="and" rules_run="22" fail_count="1" pass_count="21" outcome="fail">
      <description>This policy tests FFV1 Matroska and Quicktime files made using vrecord and checks their validity</description>
      <rule name="Is the video format FFV1?" value="Format" tracktype="Video" occurrence="*" operator="=" xpath="mi:MediaInfo/mi:track[@type='Video'][*]/mi:Format='FFV1'" outcome="pass"/>
      <rule name="FFV1 is version 3.4 or later?" value="Format_Version" tracktype="Video" occurrence="*" operator="&gt;=" xpath="mi:MediaInfo/mi:track[@type='Video'][*]/mi:Format_Version&gt;='3.4'" outcome="pass"/>
      <rule name="FFV1 is encoded in GOP size of 1?" value="Format_Settings_GOP" tracktype="Video" occurrence="*" operator="=" xpath="mi:MediaInfo/mi:track[@type='Video'][*]/mi:Format_Settings_GOP='N=1'" outcome="pass"/>
[...]

PR #350, opened 2018-06-28
use alternate mediainfo expressions for BT.470 B/G

dericed:
syncing to https://github.com/MediaArea/MediaInfoLib/commit/c75a5f3a864c92ee472b706ba362677f5f0a4697

addresses https://github.com/amiaopensource/vrecord/issues/349


JeromeMartinez:
Sorry about the regression, definitely need to find a better method for changing some display. Maybe use a "OR" in case the policy is used with an older version.

dericed:
good point @JeromeMartinez. AFK just now, but this pr should be updated to support either expression.

bturkus:
tested, and looks good. thanks for the quick fix!

dericed:
@bturkus, this is updated to @JeromeMartinez's suggestion, please re-review.

PR #352, opened 2018-07-02
Alt input

dericed:


vrecord
```diff
@@ -1165,16 +1168,20 @@ MIDDLEOPTIONS+=(${EXTRAOUTPUTOPTIONS[@]})
 # set up input and playback
 _set_up_drawtext
 _get_decklink_inputs
```
privatezero: This needs to be moved under the `else` at 1173. After making that change this works on my mac laptop!

privatezero: (and by this I mean the `_get_decklink_inputs`

dericed:
thx @privatezero, please re-review

Issue #282, opened 2018-03-15
[wish] other platforms than macOS


retokromer:
Most of `vrecord` can be ported quite easily to other OS as well. And, obviously, I’m thinking about the various flavours of Linux (including the Linux subsystem for Windows). The easiest way to do this by «just» skipping the GUI and use the CLI alone.

Some recently added parts would benefit from a more global approach. If we decide to be less Apple centric, a few parts need to be adapted. For example [this line](https://github.com/amiaopensource/vrecord/blob/6031016cc15a625202bced7eb7d25eb24fe120f9/qcview.lua#L22) is macOS specific, but the actual value could be assigned depending on the running OS.

I’m also thinking about switching from `Pashua` to an implementation with `Qt`. Any thoughts? Thank you very much!

privatezero:
I also am interested in this! I have experimented successfully with running elements of vrecord in linux, and am very interested in a version that would be manageable for Linux subsystem in Windows.

Along with things like font paths, I think we would have to consider some of the different possible install situations, with vrecord in its current form being very tied to Homebrew for a functional install.

I have found that Linuxbrew works very well, but not as seamlessly as Homebrew which could cause some problems for parts of vrecord that assume a Homebrew install.

retokromer:
Thank you, @privatezero! FYI: About one year ago, it was possible to use a formula for version 0.7.28 which was working on both Homebrew and Linuxbrew. I don’t think that the improvements made during the last year will compromise this.

retokromer:
I suggest to:

- add a test that the program is running on a supported OS
- modify the configuration with a case:
```Bash
case "$(uname -s)" in
    Darwin)
        DEFAULTFONT="/Library/Fonts/Andale Mono.ttf"
        # etc.
        ;;
    Linux)
        # etc.
        ;;
esac
```
- modify when needed with an if:
```Bash
if [[ "$(uname -s)" = "Darwin" ]]; then
    # Pashua stuff
fi
```

The first official non-macOS version could skip the `Pashua` GUI and work under the strict CLI, while waiting for a `Qt` implementation in the second version.

retokromer:
FYI: For my coming Asian workshops, I’m planning to publish an alternative formula, which patches the current release as needed, in order to get an installation working as well on various Linuxes, including via Windows 10. This could then be considered as a proof of concept, and allow us to decide if we wish to become less macOS-centric than now or not.

retokromer:
[...]

PR #353, opened 2018-07-03
Refactor conf and lookups

dericed:
needs testing but these two commits simplify the construction of the pashua configuration file and consolidation the functions that interepret the meaning of the user options.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -572,54 +572,39 @@ _writeingestlog(){
     fi
 }
 
-# set vrecord options as specified by user
-_lookup_video_input(){
+# decipher vrecord options as specified by user
+_lookup_choice(){
+    case "${2}" in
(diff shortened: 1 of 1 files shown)


privatezero:
Seems to be functioning as intended! Anything specific you wanted tested out?

dericed:
I guess i want to make sure that the added complexity is worth the line count reduction. This does reduce the need to repeat arrays of optional repeated though.

privatezero:
cool! I'll play around with it a bit more on the internal end of things to see if I notice anything else

retokromer:
>This does reduce the need to repeat arrays of optional repeated though.

That’s a good move, in my opinion (also in the optic of https://github.com/amiaopensource/vrecord/issues/282)!

PR #354, opened 2018-07-05
more refactoring

dericed:
over 10% line reduction, some edge-case bug fixes, reduction in redunancy

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1062,7 +1062,7 @@ if [[ -z "${TECHNICIAN}" ]] ; then
     read TECHNICIAN
 fi
 
-_report -d "Summary: ${PIXEL_FORMAT} bit ${CODECNAME} ${CONTAINER_CHOICE} file from ${STANDARD_CHOICE} ${VIDEO_INPUT_CHOICE} ${AUDIO_INPUT_CHOICE}. Frame MD5s=${FRAMEMD5_CHOICE}, QCTools XML=${QCTOOLSXML_CHOICE}, and Technician=${TECHNICIAN}. Inputs recorded to ${DIR}/${ID}${SUFFIX}.${EXTENSION} and Auxiliary Files created in ${LOGDIR}"
+_report -d "Summary: ${PIXEL_FORMAT} bit ${CODECNAME} ${CONTAINER_CHOICE} file from ${STANDARD_CHOICE} ${VIDEO_INPUT_CHOICE} ${AUDIO_INPUT_CHOICE}. Frame MD5s=${FRAMEMD5_CHOICE}, QCTools XML=${QCTOOLSXML_CHOICE}, and Technician=${TECHNICIAN}. Inputs recorded to ${VRECORD_OUTPUT} and Auxiliary Files created in ${LOGDIR}"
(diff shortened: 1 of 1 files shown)


privatezero:
On first test there seems to be some problem with some setting of variables in GUI vs CLI - I set the QCTools option in GUI to `Undeclared`, which causes the CLI to prompt as expected. However, echoing out the variable later in the process shows that `"${QCTOOLSXML_CHOICE}"` has not been updated to the value selected in GUI and remains `Undeclared`. I'll look a bit more at it to see why this is the case!

privatezero:
Another thing - when I enter an erroneous selection via the CLI (such as if there are 3 choices and I enter 4) it throws the error notification that I made an invalid selection, but it then goes on to the next option anyway instead of looping back until I enter a valid choice.

retokromer:
I confirm the reported issues with the CLI. We are testing here in anticipation of https://github.com/amiaopensource/vrecord/issues/282#issuecomment-397864416 as well.

privatezero:
I assume that a minor tweak to these changes would probably fix these issues, but I wasn't able to track it down today - felt like I was close but ran out of time before having to focus on other work! I'll be a bit pressed for time due to travel over the next few days but will try to take a look again as needed!

dericed:
Hi @privatezero, thx for reviewing. For both issues https://github.com/amiaopensource/vrecord/pull/354/commits/3646e9414a39597e7a208d620e9994d639d31143 was updated. I added the `declare` line so that re-requested options are stored as expected. The issue with not re-asking after a bad option was because I originally had:

```
         select OPTION in "${CHOICES[@]}" ; do
             _lookup_choice "${OPTION}" "${REPLY}"
+        LOG_OF_OPTIONS+="${VALUE_NAME}: ${OPTION}\n"
             [[ "${?}" -eq 0 ]] && break
         done
```

So the `$?` was noting the array assignment rather than the error in the _lookup_choice. The array assignment is moved now.

privatezero:
Just tested - it threw an error, but changing line 875
`declare "${VALUE_NAME}" "${OPTION}"`
to
`declare "${VALUE_NAME}"="${OPTION}"` fixed it

privatezero:
On my end setting the `QCTOOLSXML_CHOICE` variable via CLI still results in it staying as 'Undeclared'. Still investigating!

privatezero:
Looks to me according to this [man](https://ss64.com/bash/declare.html) that declare creates a local variable within a function. There is an option to add `-g` flag for global in newer versions of bash, but  looks like this isn't present in the one installed in macOS. Might have to explicitly set those variables (the ones listed under #other) to themselves in order to use `declare`

privatezero:
on further messing around it seems that (on my end) changing line 865 to
`export "${VALUE_NAME}"="${OPTION}"`
fixes both issues. Any possible negative effects of this that I didn't think of?

dericed:
@privatezero I updated the pr with your advise. Please re-review.

privatezero:
On a trip to WSU Vancouver so am away from my mac for a bit, but it was working on my end just with that change so gonna assume that still holds!

Issue #339, opened 2018-06-18
CONTRIBUTING.md


retokromer:
Should [this file](https://github.com/amiaopensource/vrecord/blob/master/Resources/Documentation/CONTRIBUTING.md) not return at the first level, in order to work automatically in GitHub?

PR #355, opened 2018-07-09
move "contributing" back to first level

retokromer:
should resolve https://github.com/amiaopensource/vrecord/issues/339


dericed:
thx

PR #357, opened 2018-07-11
 png of QCTools data

nkaplan4250:
The script uses gnuplot to produce a png of the QCTools data as a means
of quickly assessing a recording. It is still in progress. There are
some issues setting the range of with the y-axis and palette gradient
in a way that is meaningful and will change to reflect the recording’s
specifications, the x-axis is just in seconds, and the position of the
plot’s keys are not quite right.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -576,11 +576,12 @@ _writeingestlog(){
 _qcgraphimage(){
     # get audio data
     qctools_CONFIGFILE_A="$(/usr/bin/mktemp /tmp/astats_A_XXXXXXXXX).csv"
-    gzcat "$1" | perl -nle 'print if not m{lavfi.(?!astats.1.Min_level|astats.1.Max_level|astats.1.Peak_level)}' | xml select -t -m "//ffprobe:ffprobe/frames/frame[@media_type='audio']" \
+    gzcat "$1" | perl -nle 'print if not m{lavfi.(?!astats.Overall.Min_level|astats.Overall.Max_level|astats.Overall.Peak_level|aphasemeter.phase)}' | xml select -t -m "//ffprobe:ffprobe/frames/frame[@media_type='audio']" \
        -v "@pkt_pts_time" -o " " \
-       -v "tag[@key='lavfi.astats.1.Max_level']/@value" -o " " \
-       -v "tag[@key='lavfi.astats.1.Min_level']/@value" -o " " \
-       -v "tag[@key='lavfi.astats.1.Peak_level']/@value" -n > "${qctools_CONFIGFILE_A}"
+       -v "tag[@key='lavfi.astats.Overall.Max_level']/@value" -o " " \
+       -v "tag[@key='lavfi.astats.Overall.Min_level']/@value" -o " " \
+       -v "tag[@key='lavfi.astats.Overall.Peak_level']/@value" -o " " \
+       -v "tag[@key='lavfi.aphasemeter.phase']/@value" -n > "${qctools_CONFIGFILE_A}"
     # get video data
     qctools_CONFIGFILE_V="$(/usr/bin/mktemp /tmp/signalstats_V_XXXXXXXXX).csv"
     gzcat "$1" | perl -nle 'print if not m{lavfi.(?!signalstats.BRNG|signalstats.TOUT|signalstats.SATMAX|signalstats.SATHIGH|signalstats.SATLOW|signalstats.SATMIN|ssim.Y|ssim.U|ssim.V)}' | xml select -t -m "//ffprobe:ffprobe/frames/frame[@media_type='video']" \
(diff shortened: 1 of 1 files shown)


nkaplan4250:
Here is an example of what it looks like: 
qc_output_graphs

dericed:
I suggest this be considered to supersede @privatezero's work at https://github.com/amiaopensource/vrecord/pull/262.

dericed:
Some comments:
- the peak level color scale should be green-red, rather than red-green-red. For ranges perhaps:
  -  dark green: the input level is less than -50 dBFS
  - light green: the input level is between -50 and -20 dBFS
  - yellow: the input level is between -20 and -9 dBFS
  - orange: the input level is between -9 dBFS and -0.5 dBFS
  - red: the input level is larger than -0.5 dBFS
  - and have a filled curve rather than a line for readability

- the graphs should be aligned by time
- add the filename to the header
- add an indicator to say that the x-axis is in seconds or display the xaxis marks in HH:MM:SS format

vrecord
```diff
@@ -572,6 +572,71 @@ _writeingestlog(){
     fi
 }
 
+# create a png of qc data graphs for quick assessment
+_qcgraphimage(){
+    # get audio data
+    qctools_CONFIGFILE_A="$(/usr/bin/mktemp /tmp/astats_A_XXXXXXXXX).csv"
+    gzcat "$1" | perl -nle 'print if not m{lavfi.(?!astats.1.Min_level|astats.1.Max_level|astats.1.Peak_level)}' | xml select -t -m "//ffprobe:ffprobe/frames/frame[@media_type='audio']" \
```
dericed: noting that this uses the first audio channel rather than the overall audio. Perhaps `astats.Overall` would be better throughout rather than `astats.1`.

vrecord
```diff
@@ -1286,6 +1351,8 @@ if [[ "${QCTOOLSXML_CHOICE}" != "No" ]] ; then
     if [[ "${BRNG_OUTLIERS}" -gt "${BRNG_OUTLIER_THRSHLD}" ]] ; then
         cowsay "$(_report -w "WARNING: Your video file contains ${BRNG_OUTLIERS} frames with pixels out of broadcast range.")"
     fi
+    _report -d "Vrecord is generating graphs from the QCTools data. One moment please."
+    _qcgraphimage "${QCXML}"
```
privatezero: I might move this above the `else` at 1342 so that the function is run only on successful confirmation of `${QCXML}`

privatezero:
I agree with the other suggestions and that this should supersede https://github.com/amiaopensource/vrecord/pull/262!

nkaplan4250:
Thank you for the suggestions! In addition to those changes the image is now a jpeg to accommodate the need for multiple color ranges. A plot of the audio phase  data has been added. The yaxis scales are set for the audio levels and broadcast range. The y-axis and the color gradient for the saturation will scale to match the video’s bit depth. The title changes to match the recording’s identifier. Below are some examples:
8bit2graphtestmat_qc_output_graphs

10bitgraphtest_qc_output_graphs

vrecord
```diff
@@ -572,6 +572,110 @@ _writeingestlog(){
...
+       -v "tag[@key='lavfi.signalstats.SATLOW']/@value" -o " " \
+       -v "tag[@key='lavfi.signalstats.SATMIN']/@value" -o " " \
+       -v "tag[@key='lavfi.ssim.Y']/@value" -o " " \
+       -v "tag[@key='lavfi.ssim.U']/@value" -o " " \
+       -v "tag[@key='lavfi.ssim.V']/@value" -n > "${qctools_CONFIGFILE_V}"
+
+    # determine SATURATION scale
+    VBitdepth="$(gzcat "$1" | perl -nle 'print if not m{<frame|</frame|<tag}' | xml sel -t -v "//stream[@codec_type='video']/@pix_fmt" -n
+    )"
+        if [[ "${VBitdepth}" = "uyvy422" ]] || [[ "${VBitdepth}" = "yuv422p" ]] ; then
```
dericed: this `if/fi` section is indented one indent more than needed

vrecord
```diff
@@ -572,6 +572,110 @@ _writeingestlog(){
...
+       -v "tag[@key='lavfi.astats.Overall.Peak_level']/@value" -o " " \
+       -v "tag[@key='lavfi.aphasemeter.phase']/@value" -n > "${qctools_CONFIGFILE_A}"
+    # get video data
+    qctools_CONFIGFILE_V="$(/usr/bin/mktemp /tmp/signalstats_V_XXXXXXXXX).csv"
+    gzcat "$1" | perl -nle 'print if not m{lavfi.(?!signalstats.BRNG|signalstats.TOUT|signalstats.SATMAX|signalstats.SATHIGH|signalstats.SATLOW|signalstats.SATMIN|ssim.Y|ssim.U|ssim.V)}' | xml select -t -m "//ffprobe:ffprobe/frames/frame[@media_type='video']" \
+       -v "@pkt_pts_time" -o " " \
+       -v "tag[@key='lavfi.signalstats.TOUT']/@value" -o " " \
+       -v "tag[@key='lavfi.signalstats.BRNG']/@value" -o " " \
+       -v "tag[@key='lavfi.signalstats.SATMAX']/@value" -o " " \
+       -v "tag[@key='lavfi.signalstats.SATHIGH']/@value" -o " " \
```
dericed: SATHIGH and a few other SAT values aren't used below so they could be removed here and from the perl parsing above to save some time

privatezero:
Just tested this and got the following error:
image

The jpeg was generated, and is lacking the phase section but otherwise looks great!

dericed:
Hi @privatezero, can you check your QCTools preferences and confirm that aphasemeter is unchecked? If so, I think either vrecord should force a certain set of qc filters or test for the presence of a filter's data before plotting it.

privatezero:
Wasn't able to confirm settings due to a charging issue, but I am willing to bet that aphase is not selected in the QCTools config. I think that forcing a standard set of filters as @dericed suggested probably makes the most sense to keep things all self contained and standardized.

PR #363, opened 2018-07-25
delete unneeded escape

retokromer:


retokromer:
>I squashed with #364 since same commit message and intent.

Thank you! I could not do it myself, but that was exactly my intent.

Issue #365, opened 2018-07-29
[wish] add information to config file


retokromer:
In order to simplify debugging, I suggest to write into the `${CONFIG_FILE}` also the available information about the used configuration (hardware and software).

karlfitzke:
Cornell AVPL would be jazzed about that, Reto.   Would it be worth having a separate file for such info (so one could independently manage each without messing up the other)?  Perhaps a "vrecord -s" (s for signal path) mode allowing us to input all this data with pull downs (or other given presets to minmize error/maximize consistency) in a similar fashion to vrecord -e?

Deck make/model/serial num; 
Deck speed
Deck to TBC connection type (e.g. S-Video); 
TBC make/model/serial num; 
TBD to ADC connection type (e.g. component); 
ADC make/model serial num;
ADC driver (or is that already in some log file?)
Computer make/model/serial num; 
OS version;
Capture Drive path (and maybe even make/model)?
Notes field (256? char)

What did I leave out?

By the way, we are just starting to use vrecord here the past few days AND LOVING IT - thanks team!  Will talk about possibility of funding definition and actual work on the above with Tre.

retokromer:
@kieranjol What’s your preference: in the same or in two different files?

kieranjol:
I'll have to check tomorrow when at work.i will need to remind myself of what kinds log logs vrecird produces at the moment

retokromer:
Thank you, that would be useful indeed! I personally think that all in one is simpler, but I can also try to make a PR for two distinct. And please let me know which information I should include.

Initially I was thinking it just for simplifying debugging, because we encounter only very little of the problems others have, but of course it can be more detailed and for archiving as well.

karlfitzke:
One file is fine.  I defer to those more in the know.  Converter firmware version may be useful to add too.  I forgot that the other day.  And "TBD to ADC" above is a typo, and should be "TBC to ADC".

kieranjol:
Aye, one file would be fine. Am I correct in thinking that this information would be automatically appended to the `_capture_options.log` file?

-K

PR #376, opened 2018-09-06
log info about capture device

dericed:
uses diskutil to gather data

partly addresses https://github.com/amiaopensource/vrecord/issues/365#issuecomment-413977518


retokromer:
I suggest to mention the dependance in the [doc](https://github.com/amiaopensource/vrecord#summary) as well.

PR #377, opened 2018-09-07
Log decklink info

dericed:
also partly resolves https://github.com/amiaopensource/vrecord/issues/365#issuecomment-413977518


kieranjol:
Wow,this is very cool that it's all automated.

PR #380, opened 2018-09-14
fix handling milliseconds with leading zero

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/379


kieranjol:
I can't test right now but thanks for fixing it!

PR #381, opened 2018-09-18
add some timecode support

dericed:
write timecode data into some file types and shows timecode metadata while in passthrough mode

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1051,9 +1051,9 @@ else
 fi
 PIPE_DECKLINK=(-c copy)
 PIPE_DECKLINK+=(-map 0)
-PIPE_DECKLINK+=(-f nut)
-PIPE_DECKLINK+=(-syncpoints none -f_strict experimental)
-PIPE_DECKLINK+=(-write_index 0)
+PIPE_DECKLINK+=(-f matroska)
+PIPE_DECKLINK+=(-write_crc32 0)
+PIPE_DECKLINK+=(-live true)
 PIPE_DECKLINK+=(-)
 WINDOW_NAME="mode:${RUNTYPE} - video:'${VIDEO_INPUT}' audio:'${AUDIO_INPUT}' - to end recording press q, esc, or close video window"
 
--- a/vrecord
+++ b/vrecord
@@ -20,6 +20,7 @@ unset DURATION
 unset TECHNICIAN
 RUNTYPE="record"
 CAPTURELOGSUFFIX="_ffmpeg_decklink_input.log"
+TIMECODELOGSUFFIX="_frame_timecodes.txt"
 unset PASHUAINSTALL
 unset EXTRAOUTPUTS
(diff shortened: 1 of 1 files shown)


kieranjol:
Woahhhhh....

bturkus:
Hey Dave,

First, just wanted to say this is the best! Thanks for making it happen! We've done a bunch of testing over the past few days, and I wanted to report our initial findings. We've had success in a variety of ways, but some combination of BM bullshit and set ups that weren't really designed for timecode capture have prevented us from making every permutation work.

My first, and probably the easiest question: is this how terminal should look during timecode capture? I'm assuming yes, but wanted to double check.

screen shot 2018-09-20 at 11 03 24 am

Another relatively easy question: how exactly is the timecode being stored in the MKV? I think I must've missed a CELLAR discussion or something, but so far I've been using MediaInfo "Time code of first frame" or mkvinfo "Tags" to figure out if things have been working properly. I'd love to get a better sense of how this is all working on element level, so if you could recommend a place where I could read more I'd appreciate it.

So...what's worked/hasn't (this is gonna get a little complicated; apologies)? We've had success with BM Decklink Duo 2 cards, but so far we haven't been able to grab any timecode with our Teranex 2Ds. I called BM tech support was told that, if the firmware is beyond version 2.1, the 2Ds should be able to capture VITC (but not LTC) and the 3Ds should be able to capture both VITC and LTC (if and only if the LTC is sent in via XLR). Either way, we've had no luck thus far, and I was hoping someone else (maybe @NMAAHC) could check it out.

With the Duo 2s, we've had mixed success. For Betacam/Digi played back in a Sony DVW-M2000, we've been able to capture rp188ltc but not vitc. We're still hunting down the best tapes for testing, but the ones we've been using seem to have both LTC and VITC. Which raises another question: would it be possible (in the future) to capture both forms of timecode at the same time? I was reading somewhere that tapes might have a screwed up LTC track but continuous VITC, and dual capture could be very AS-07. And though we weren't able to get VITC with the Betacams, we did manage to capture it with a DVCam (DSR-2000), going out SDI.

One general thought: I don't know how others would feel, but I think it'd be great to have timecode displayed on screen during capture, and not just during passthrough. But I guess I could understand if others would disagree.

Again, this is such great stuff. More testing, more people!

Ben

kieranjol:
+1 timecode displayed on screen

retokromer:
@dericed Could you not silence the NUT errors? (Untested!)

dericed:
The warnings in https://github.com/amiaopensource/vrecord/pull/381#issuecomment-423331645 were addressed in https://github.com/amiaopensource/vrecord/pull/381/commits/c4dd56a55e2ca92284fbec0b5ac66f5c5aed527c. Ready for re-review. I suggest to delay merge until after ffmpegdecklink is updated to 4.1.

PR #387, opened 2018-10-01
remove redundant code

retokromer:
The `else` includes the `previous elif [[ "${db}" = 1 ]]` as well.


retokromer:
ping

dericed:
I remember intentionally adding these lines because of a plan to distinguish what happens next based upon this behavior, but at the moment I don't remember what the plan was. I suppose I'm +/-0 on this pr.

retokromer:
The goal is to simplify the code, in order to find the part which now slows down the processing.

retokromer:
@dericed Could you find the reason for this «dirty» code?

retokromer:
ping

dericed:
I’m not in favor of this commit. For me it doesn’t seem simpler since the patch removes the only usage of the variable from a button that is intentionally defined. The design of the form makes a bit less sense with this approach and for no functional benefit. Also the distinction between how the form is closed is lost. @weevz, any comment here?

PR #398, opened 2018-11-27
separate views for passthrough and recording

dericed:


libbyhopfauf:
I will look at this as soon as I am back in Seattle!! Thank you, @dericed!!

bturkus:
just reviewed this one, and the ability to set diff views for passthrough and capture seems to working well. one thing I noticed: it looks like the key-bindings/lua part of mpv isn't working in this new version. could someone else confirm?

PR #405, opened 2018-12-13
Conditionals for system

privatezero:
This isn't finished yet, but does allow for some more testing. As it already has some conflicts gonna open a PR for review before it gets any more divergent.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -24,7 +24,6 @@ CAPTURELOGSUFFIX="_ffmpeg_decklink_input.log"
 TIMECODELOGSUFFIX="_frame_timecodes.txt"
 unset EXTRAOUTPUTS
 UNSELECTEDOPTION="unselected"
-DEFAULTFONT="/Library/Fonts/Andale Mono.ttf"
 SAT_OUTLIER_THRSHLD=14
 AUD_OUTLIER_THRSHLD=10
 BRNG_OUTLIER_THRSHLD=14
@@ -38,21 +37,21 @@ MPVOPTS+=(--script "${SCRIPTDIR}/qcview.lua")
 MPVOPTS+=(--really-quiet)
 
 # Set variables for system differences
-SYSTEM_TEST="$(uname -s)"
-if [[ "${SYSTEM_TEST}" = "Darwin" ]] ; then
+QUERY_OS_TYPE="$(uname -s)"
+if [[ "${QUERY_OS_TYPE}" = "Darwin" ]] ; then
     OS_TYPE="macOS"
-    DEFAULTFONT='/Library/Fonts/Andale Mono.ttf'
+    DEFAULTFONT="/Library/Fonts/Andale Mono.ttf"
     HARDWARE_REPORT=$(system_profiler SPHardwareDataType)
     CORE_COUNT="$(_parse_report "Total Number of Cores" "${HARDWARE_REPORT}")"
     OPEN_COMMAND="open"
-elif [[ "${SYSTEM_TEST}" = "Linux" ]] ; then
+elif [[ "${QUERY_OS_TYPE}" = "Linux" ]] ; then
     OS_TYPE="linux"
-    DEFAULTFONT='/usr/share/fonts/truetype/liberation/LiberationSans-Regular.ttf'
+    DEFAULTFONT="/usr/share/fonts/truetype/liberation/LiberationSans-Regular.ttf"
     HARDWARE_REPORT=$(lscpu)
     CORE_COUNT="$(nproc --all)"
     OPEN_COMMAND="xdg-open"
 else
-    echo "Unsupported OS, ${SYSTEM_TEST}, detected. Exiting."
+    echo "Unsupported OS, ${QUERY_OS_TYPE}, detected. Exiting."
     exit 1
 fi
 
--- a/vrecord
+++ b/vrecord
@@ -38,20 +38,21 @@ MPVOPTS+=(--script "${SCRIPTDIR}/qcview.lua")
 MPVOPTS+=(--really-quiet)
 
 # Set variables for system differences
-if [[ "$(uname -s)" = "Darwin" ]] ; then
+SYSTEM_TEST="$(uname -s)"
+if [[ "${SYSTEM_TEST}" = "Darwin" ]] ; then
     OS_TYPE="macOS"
     DEFAULTFONT='/Library/Fonts/Andale Mono.ttf'
     HARDWARE_REPORT=$(system_profiler SPHardwareDataType)
     CORE_COUNT="$(_parse_report "Total Number of Cores" "${HARDWARE_REPORT}")"
     OPEN_COMMAND="open"
-elif [[ "$(uname -s)" = "Linux" ]] ; then
+elif [[ "${SYSTEM_TEST}" = "Linux" ]] ; then
     OS_TYPE="linux"
     DEFAULTFONT='/usr/share/fonts/truetype/liberation/LiberationSans-Regular.ttf'
     HARDWARE_REPORT=$(lscpu)
     CORE_COUNT="$(nproc --all)"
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -1183,33 +1196,37 @@ QCXML="${LOGDIR}/${ID}${SUFFIX}.${EXTENSION}.qctools.xml.gz"
 touch "${INGESTLOG}"
 _writeingestlog "computer_name" "$(uname -n)"
 _writeingestlog "computer_model_name" "$(_parse_report "Model Name" "${HARDWARE_REPORT}")"
-_writeingestlog "computer_model_id" "$(_parse_report "Model Identifier" "${HARDWARE_REPORT}")"
-_writeingestlog "computer_processor_name" "$(_parse_report "Processor Name" "${HARDWARE_REPORT}")"
-_writeingestlog "computer_processor_speed" "$(_parse_report "Processor Speed" "${HARDWARE_REPORT}")"
-_writeingestlog "computer_processor_count" "$(_parse_report "Number of Processors" "${HARDWARE_REPORT}")"
+if [[ OS_TYPE = "macOS" ]] ; then
```
dericed: this test will always be false, maybe a quoted variable :)

vrecord
```diff
@@ -1183,33 +1196,37 @@ QCXML="${LOGDIR}/${ID}${SUFFIX}.${EXTENSION}.qctools.xml.gz"
...
-_writeingestlog "capture_device_protocol" "$(_parse_report "Protocol" "${CAPTURE_DEVICE_INFO}")"
-_writeingestlog "capture_device_name" "$(_parse_report "Device / Media Name" "${CAPTURE_DEVICE_INFO}")"
-_writeingestlog "capture_device_block_size" "$(_parse_report "Device Block Size" "${CAPTURE_DEVICE_INFO}")"
-_writeingestlog "capture_device_location" "$(_parse_report "Device Location" "${CAPTURE_DEVICE_INFO}")"
-_writeingestlog "capture_device_partition_type" "$(_parse_report "Partition Type" "${VOLUME_INFO}")"
-_writeingestlog "capture_device_partition_name" "$(_parse_report "Name (User Visible)" "${VOLUME_INFO}")"
-_writeingestlog "capture_device_volume_uuid" "$(_parse_report "Volume UUID" "${VOLUME_INFO}")"
-_writeingestlog "capture_device_volume_available" "$(_parse_report "Volume Available Space" "${VOLUME_INFO}")"
-_writeingestlog "capture_device_volume_total" "$(_parse_report "Volume Total Space" "${VOLUME_INFO}")"
+if [[ OS_TYPE = "macOS" ]] ; then
```
dericed: same issue mentioned before

vrecord
```diff
@@ -37,6 +37,24 @@ MPVOPTS+=(--load-scripts=no)
...
+    CORE_COUNT="$(_parse_report "Total Number of Cores" "${HARDWARE_REPORT}")"
+    OPEN_COMMAND="open"
+elif [[ "$(uname -s)" = "Linux" ]] ; then
+    OS_TYPE="linux"
+    DEFAULTFONT='/usr/share/fonts/truetype/liberation/LiberationSans-Regular.ttf'
+    HARDWARE_REPORT=$(lscpu)
+    CORE_COUNT="$(nproc --all)"
+    OPEN_COMMAND="xdg-open"
+else
+    echo 'Unsupported OS detected. Exiting.'
```
dericed: If you're going to exit here, you should include the output of `$(uname -s)` so the user gets a clue.

vrecord
privatezero: oh christ - that's what I get for blindly cut pasting myself. I think I need to up my font size in my editor or something!!

vrecord
privatezero: makes sense!

dericed:
mind resolving the conflict, you were working with the master branch before @siggaregina's update.

privatezero:
Deleted pashua chunk and should be resolved!

vrecord
```diff
@@ -37,6 +37,25 @@ MPVOPTS+=(--load-scripts=no)
 MPVOPTS+=(--script "${SCRIPTDIR}/qcview.lua")
 MPVOPTS+=(--really-quiet)
 
+# Set variables for system differences
+SYSTEM_TEST="$(uname -s)"
+if [[ "${SYSTEM_TEST}" = "Darwin" ]] ; then
```
dericed: I think the variable could be better named.

vrecord
```diff
@@ -37,6 +37,25 @@ MPVOPTS+=(--load-scripts=no)
...
+    OS_TYPE="macOS"
+    DEFAULTFONT='/Library/Fonts/Andale Mono.ttf'
```
dericed: Suggest double quotes (here and below), also with this var assigned within the condition, then line 27 may be removed.

vrecord
```diff
@@ -37,6 +37,25 @@ MPVOPTS+=(--load-scripts=no)
...
+    HARDWARE_REPORT=$(system_profiler SPHardwareDataType)
+    CORE_COUNT="$(_parse_report "Total Number of Cores" "${HARDWARE_REPORT}")"
```
dericed: untested, but I suspect this would not work since you're using a function before it is defined

privatezero: yes! that would appear to be the case - thanks!

vrecord
privatezero: would you prefer something more explicit? FIND_SYSTEM_TYPE?

privatezero:
Any more changes required for this to be merged?

dericed:
sorry still testing

privatezero:
Cool! No worries - good to have rigorous testing on the mac side to make sure none of the Linux tweaks break anything!

retokromer:
Testing here as well.

retokromer:
@privatezero What is the utility of having two values `QUERY_OS_TYPE` and `OS_TYPE`? I am asking, because I did use only one…

vrecord
```diff
@@ -131,6 +130,25 @@ _check_mpv(){
...
+    OPEN_COMMAND="open"
+elif [[ "${QUERY_OS_TYPE}" = "Linux" ]] ; then
+    OS_TYPE="linux"
+    DEFAULTFONT="/usr/share/fonts/truetype/liberation/LiberationSans-Regular.ttf"
+    HARDWARE_REPORT=$(lscpu)
+    CORE_COUNT="$(nproc --all)"
+    OPEN_COMMAND="xdg-open"
+else
+    echo "Unsupported OS, ${QUERY_OS_TYPE}, detected. Exiting."
```
retokromer: To me, that way it would be easier to read: 
`echo "Unsupported OS detected: ${QUERY_OS_TYPE}. Exiting."`

privatezero:
@retokromer I considered doing it all with one variable, but I thought that with the output of the system check not always equaling the 'human readable' common name of systems (for example `Darwin` for `macOS`) that it might be a good idea to parse it into a second variable that matched with common names.

My logic was that this would make it more extensible in the future if there was a need to forefront the system type to the user via an alert or log, even though it is less "efficient" at the moment. BUT I'm certainly not married to doing it that way if the consensus of others is to keep it all with the raw output of `uname -s`.

retokromer:
@privatezero Thank you for the answer! I was just asking, not blocking at all. Your explanation makes sense. (I grew up with 256 Bytes RAM and 128 KB external storage… therefore I’m always obsessively trying to reduce both CPU cycles and needed memory… almost 40 years later ;-)

privatezero:
thanks for testing @retokromer! This doesn't yet have all the variables to match macOS functions in vrecord (for logging etc) but it seems to be a start!

retokromer:
Indeed, @privatezero, it is a very good start! Possibly on 26<sup>th</sup> we can test as well that there are not regressions on the macOS side.

Issue #404, opened 2018-12-10
add analog_rca and analog_xlr as distinct input options


karlfitzke:
Wondering if anyone else using BlackMagic UltraStudio 4K has noticed that as soon as you hit "Return" key to record with vrecord, the XLR audio inputs are selected, regardless of how you set that up in the BlackMagic Desktop Video Setup app that we also use to set input levels.

We can of course use adapter cables, so we always input XLR audio (from RCA deck outputs).  Or to capture stronger input signal levels, I imagine we can work around with RCA to RCA cables by using the BlackMagic Video Setup app to set things back to RCA after recording begins.

Or have I gotten ahead of things here, i.e. are we doing something wrong?

We've seen this behavior with vrecord builds from 2018-08-12 and 2018-12-05.    Using SDI video input.  Matroska 10-bit ffv1 video, analog audio selected.  Thanks.   -Karl

dericed:
vrecord doesn't use the selections of the Desktop Video application. Use `vrecord -e` to get a preference pane where you can select the audio input you prefer.

karlfitzke:
Thanks for the prompt reply, Dave.  

vrecord only allows an "analog" audio input choice, which is why we tried Desktop Video to further clarify which one, RCA or XLR audio, just as we do to successfully set input level (attenuation).  Can selection of either analog RCA or analog XLR audio be added to the To Do list?

dericed:
@karlfitzke so does the BlackMagic UltraStudio 4K has both the analog rca and analog xlr inputs simultaneously on it. The other cards Ive used only have one or the other depending on the breakout cable.

karlfitzke:
Yes, the 4K has both RCA and XLR analog audio inputs - but only two channels each, and one or the other pair (RCA or XLR) selectable at any given time with the software that I have, and likely hardware too to save money and back panel space - read on if interested in hearing more. 

Only having two audio input channels of course assumes, sadly to us archival people, that no one is capturing old four channel video tapes anymore.   We use the 4K for two channel video only as a result.

[...]

PR #407, opened 2018-12-17
add analog rca and xlr options

dericed:
resolves https://github.com/amiaopensource/vrecord/issues/404

### DIFF
--- a/Resources/Documentation/settings.md
+++ b/Resources/Documentation/settings.md
@@ -16,9 +16,9 @@ All of the options in the vrecord GUI (which appears when running `vrecord -e`),
 
 **Select a recording directory** — Choose the location (on an internal or external hard drive) where you want your resulting video files and logs to be saved.
 
-**Select video input** — Choose how the video signal will be entering the capture device. You can receive the video signal through Composite, SDI, Component, or S-Video cables.
+**Select video input** — Choose how the video signal will be entering the capture device. You can receive the video signal through Composite, SDI, Component, or S-Video cables. The "Scan supported a/v inputs" button will test your capture hardware against each input type, to see what is and is not supported.
 
-**Select audio input** — Choose how the audio signal will be entering the capture device. You can receive the audio signal through Analog (such as XLR), SDI, or other digital audio cables.  
+**Select audio input** — Choose how the audio signal will be entering the capture device. You can receive the audio signal through Analog (such as XLR), SDI, or other digital audio cables. The "Scan supported a/v inputs" button will test your capture hardware against each input type, to see what is and is not supported.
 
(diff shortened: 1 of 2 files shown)


privatezero:
Shouldn't the option for generic 'analog' be removed to avoid possible confusion since the addition of more specific options basically supersedes it?

privatezero:
On that note - it might be a good idea to add the line levels in the config settings so something like:

`RCA Inputs (-10 dBv)`
`XLR Inputs (+4 dBu)`

To make things a little clearer on when people should be using which setting?

retokromer:
nit: is it `dBv` in the USA? EBU uses `dBV`.

privatezero:
No - you are correct! Should be `-10 dBV`

dericed:
@privatezero, not sure about adding static line level measurements such as -10 and +4. This would need some testing to ensure that custom line level alternations added in the Desktop Video application are not used in vrecord. For instance if RCA input default is -10 but someone adjusts this in Desktop Video and vrecord does use it, then we shouldn't report -10 as it could be misleading.

Also I'd prefer to leave the Analog option since many users are accustomed to it. And it allows switching between the breakdown cables without necessarily having to adjust the audio input setting.

privatezero:
@dericed, are you referring to the custom input gain settings in Desktop Video? I agree that would be a problem if those gain settings are passed through to vrecord without the user's knowledge (and indeed some testing should probably be done for that), but that doesn't relate the the line level settings which will always be either -10 dBV or +4 dBu depending on source.

Has there been any testing to confirm that using the generic `analog` setting within the FFmpeg device  actually does  in fact does select the correct input levels automatically depending on which breakout cable is being used?

dericed:
updated based on @privatezero feedback, but with my UltraStudio Express I get this if I select either analog_xlr or analog_rca:

```
[decklink @ 0x7fbb5c8a0200] Device does not support selected audio input.
[decklink @ 0x7fbb5c8a0200] Could not set input configuration
UltraStudio Express: Input/output error
```

I haven't found a way to detect what inputs are supported by the connected device.

dericed:
This commit is updated to include a method of testing whether or not the listed audio and video inputs are supported by the hardware. Please re-review.

retokromer:
ping

privatezero:
Somehow missed this update! I'll give it a test tonight for the new 'scan inputs' button. I can confirm reproducing the original behavior of the Blackmagic card used for testing being unable to select anything beyond the generic analog input.

dericed:
rebasing upon https://github.com/amiaopensource/vrecord/pull/489

privatezero:
Just tested - the scan inputs button works once, but upon second click seems to fire up the passthrough mode for some reason

PR #412, opened 2019-01-25
move png+xml+lua to Resources folder and call it from there or prefix…

dericed:
….install


vrecord
```diff
@@ -10,10 +10,11 @@ GTK_RECENT_DIR="${HOME}/.local/share"
 unset VERSION
 if [[ $(dirname "$(command -v "${0}")") = "/usr/local/bin" ]] ; then
     VERSION=$(TMP=$(brew info vrecord | grep ".*\*$" | grep -Eo "/vrecord/.* \(") ; echo "${TMP:9:(${#TMP}-11)}")
-    LOGO_PATH="/usr/local/Cellar/vrecord/${VERSION}/vrecord_logo.png"
+    RESOURCE_PATH="/usr/local/Cellar/vrecord/${VERSION}/Resources"
 else
-    LOGO_PATH="${SCRIPTDIR}/vrecord_logo.png"
+    RESOURCE_PATH="${SCRIPTDIR}/Resources"
```
dericed: @privatezero is LOGO_PATH still needed? Still like both this and vrecord_logo.png are unused, may I remove?

privatezero: Seems obsolete now that you are calling the source images individually! I'd say they can be removed unless you are sentimental about the good old pashua days!

PR #424, opened 2019-04-03
Update installation_and_setup.md

bturkus:
removed gnuplot, added basic linux installation

### DIFF
--- a/Resources/Documentation/installation_and_setup.md
+++ b/Resources/Documentation/installation_and_setup.md
@@ -1,5 +1,7 @@
 ## Installing vrecord
 
+### macOS
+
 If you use macOS, you can easily install vrecord using a package manager called Homebrew. To install Homebrew, follow [these instructions](https://brew.sh/).
 
 To install vrecord, run the following two commands in a [Terminal window](https://en.wikipedia.org/wiki/Terminal_%28macOS%29):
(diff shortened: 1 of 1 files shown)


Resources/Documentation/installation_and_setup.md
```diff
@@ -34,7 +29,18 @@ brew upgrade
 ```
 (This command will upgrade all of the programs you've installed through Homebrew.)
 
-Thus far installing vrecord on Linux using Linuxbrew has not been successful.
+### Linux
+
+As of version 2019-01-19, vrecord can now be installed on select Linux distributions, most easily via [Linuxbrew](https://docs.brew.sh/Homebrew-on-Linux), with the same macOS Homebrew Terminal commands:
```
privatezero: Think it isn't technically called `Linuxbrew` any more!

Resources/Documentation/installation_and_setup.md
```diff
@@ -34,7 +29,18 @@ brew upgrade
...
+```
+brew tap amiaopensource/amiaos
+brew install vrecord
+```
+But if find yourself experiencing Linux installation issues, it may be beneficial to first install vrecord's core dependencies, and then install vrecord without dependencies:
```
privatezero: maybe add a caveat about how this method will only support core functions (things like qcli will still need to be installed before they work)

Resources/Documentation/installation_and_setup.md
retokromer: Indeed, the document begins now with the sentence:

> The Homebrew package manager may be used on Linux and Windows Subsystem for Linux (WSL).

PR #426, opened 2019-04-04
Changes to suffix

privatezero:
Should address https://github.com/amiaopensource/vrecord/issues/342

Also made some changes to indentation and some tweaks for linux

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -280,203 +280,205 @@ _edit_prefs() {
                 </list>
             </vbox>
         </hbox>
-    <hbox>
-        <vbox>
-            <text>
-                <label>Embed digitization logs in video file (Matroska ONLY)</label>
-            </text>
-            <list selection-mode=\"1\" selected-row=\"$(_get_index_of_value "${EMBED_LOGS_CHOICE}" "${EMBED_LOGS_OPTIONS[@]}")\">
-                <variable>EMBED_LOGS_CHOICE</variable>
-                $(_expand_list2items "${EMBED_LOGS_OPTIONS[@]}")
-            </list>
-        </vbox>
-        <vbox>
-            <text>
-                <label>Create frame-level MD5 checksums? (recommended)</label>
-            </text>
-            <list selection-mode=\"1\" selected-row=\"$(_get_index_of_value "${FRAMEMD5_CHOICE}" "${FRAMEMD5_PPTIONS[@]}")\">
-                <variable>FRAMEMD5_CHOICE</variable>
-                $(_expand_list2items "${FRAMEMD5_OPTIONS[@]}")
-            </list>
-        </vbox>
-        <vbox>
-            <text>
-                <label>Select File Format</label>
-            </text>
-            <list selection-mode=\"1\" selected-row=\"$(_get_index_of_value "${CONTAINER_CHOICE}" "${CONTAINER_OPTIONS[@]}")\">
-                <variable>CONTAINER_CHOICE</variable>
-                $(_expand_list2items "${CONTAINER_OPTIONS[@]}")
-            </list>
-        </vbox>
-        <vbox>
-            <text>
-                <label>Select Video Codec</label>
-            </text>
-            <list selection-mode=\"1\" selected-row=\"$(_get_index_of_value "${VIDEO_CODEC_CHOICE}" "${VIDEO_CODEC_OPTIONS[@]}")\">
-                <variable>VIDEO_CODEC_CHOICE</variable>
-                $(_expand_list2items "${VIDEO_CODEC_OPTIONS[@]}")
-            </list>
-        </vbox>
-        <vbox>
-            <text>
-                <label>FFV1 Slice Count</label>
-            </text>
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -406,6 +406,24 @@ _edit_prefs() {
...
+                <default>\"${PREFIX}\"</default>
+                <variable>PREFIX</variable>
+            </entry>
+        </vbox>
+        <vbox>
+            <text>
+                <label>Enter an optional file suffix</label>
+            </text>
+            <entry>
+                <default>\"${USER_SUFFIX}\"</default>
```
dericed: why $PREFIX vs $USER_SUFFIX?

vrecord
```diff
@@ -1298,8 +1318,15 @@ if [[ "${VIDEO_CODEC_CHOICE}" = "FFV1" ]] ; then
 fi
 _review_option "AUDIO_CODEC_CHOICE" "Which audio codec?" "${AUDIO_CODEC_OPTIONS[@]}"
 
+# Check for user suffix to override automatic settings
+
+if [[ -n "${USER_SUFFIX}" ]] ; then
+    SUFFIX="${USER_SUFFIX}"
```
dericed: hmm, then there's no way to disable the suffix. Perhaps it's worth removing the auto-suffixing and then in _verify_conf_items add
```
if [[ "$(grep -c "${CONF_VARIABLE}" "${CONF_FILE}")" = "0" ]] ; then
    if [[ "${CONF_VARIABLE}" = "USER_SUFFIX" ]] ; then
        _report -w "Please note that moving forward, vrecord will not automatically add suffixes. If you'd like a suffix, please indicate it in the new section of the preferences.
    fi
   echo "${CONF_VARIABLE}=\"\"" >> "${CONF_FILE}"
fi
```

and then call `_verify_conf_items "USER_SUFFIX" "${CONFIG_FILE}"` later in the script

vrecord
```diff
@@ -1367,8 +1368,8 @@ _report -q "Hit enter to start recording"
 read
 
 # create log of vrecord decisions
-INGESTLOG="${LOGDIR}/${ID}_capture_options.log"
-QCXML="${LOGDIR}/${ID}${SUFFIX}.${EXTENSION}.qctools.xml.gz"
+INGESTLOG="${LOGDIR}/${PREFIX}${ID}${SUFFIX}_capture_options.log"
+QCXML="${LOGDIR}/${PREFIX}${ID}${SUFFIX}.${EXTENSION}.qctools.xml.gz"
```
dericed: why not re-use $FULL_OUTPUT_ID here.

dericed:
Maybe the prefix/suffix should be added to the output gnuplot jpeg as well. Currently the filename of the jpeg and the title written within it do not include them. Same with the "_vitc_frame_timecodes.txt" output.

privatezero:
don't have time to run a test rn, but think that should do it!

vrecord
```diff
@@ -1130,7 +1160,7 @@ while [[ "${RUNTYPE}" = "edit" ]] ; do
     _edit_mode
 done
 
-if [[ "${INVERT_PHASE}" = 1 ]] ; then
+if [[ "${INVERT_PHASE}" = 'true' ]] ; then
```
retokromer: Why single quotation marks (`'`) and not double quotation marks (`"`) here?

vrecord
```diff
@@ -1298,8 +1328,18 @@ if [[ "${VIDEO_CODEC_CHOICE}" = "FFV1" ]] ; then
 fi
 _review_option "AUDIO_CODEC_CHOICE" "Which audio codec?" "${AUDIO_CODEC_OPTIONS[@]}"
 
+# Check for user suffix to override automatic settings
 
-VRECORD_OUTPUT="${DIR}/${ID}${SUFFIX}.${EXTENSION}"
+if [[ -n "${USER_SUFFIX}" ]] ; then
+    SUFFIX="${USER_SUFFIX}"
+elif [[ "${NO_SUFFIX}" = 'true' ]] ; then
```
retokromer: ditto

vrecord
dericed: There's no content to interpolate so `'true'` and `"true"` would be identical in this case.

dericed: In the script, there are 3 conditions to test the value of $INVERT_PHASE, but only 2 are changed, so the commit only partly fixes the issue.

retokromer: >There's no content to interpolate so 'true' and "true" would be identical in this case.

I know that (and in past I used to distinguish this myself!), but I was told that today it’s considered good practice to use only double quotation marks in Bash. We have many other places in the code where no interpolation occurs.

dericed: I don't want to block the merge, so I'll stall the merge for a bit to allow the contributor to have a chance to consider, else consider a subsequent pr.

retokromer: Oh, my remark it’s not blocking at all! I was just wondering why here.

privatezero: Style-wise I prefer to use `'` in situations like this where there is no interpolation as I feel that it helps distinguish in the code where things like variable substitution is actually occurring, as well as forcing me to be more conscious about differences between quotes in their different applications (such as quotes in quotes).

retokromer: Then I will open a PR to systematise the different use of `'` and `"`.

Issue #434, opened 2019-05-30
avectorscope in qc view?


bturkus:
does anyone have a sense of how difficult it would be to add avectorscope to vrecord's views? would it be easiest to do it in qcview lua? and has this been considered before? if so, and it's been set aside for a good reason, my bad.

dericed:
iirc there's some issue with syncing audio and video output in ffplay when audio filters are used in the filter to produce visualizations, i'll take a look and see if its still an issue

retokromer:
Do we not use mpv?

privatezero:
I def think that _shouldn't_ be too hard to implement in qc view and possibly as a replacement for part of the window in the audio passthrough mode (which also uses mpv). When I worked on trying to get the basic audio bars into vrecord, FFplay was not up to the job for some reason, with the combination of video and audio filtering massively contributing to lag vs the source material (and the inevitable buffer over runs and a gazillion dropped frames).

bturkus:
I'd def be open to anyone else doing this, but if I gotta struggle thru it I will. This the line that would need tweaking, right? 

https://github.com/amiaopensource/vrecord/blob/d79a79f56a50c7245ec9e9c8c6a83bb91af579ea/Resources/qcview.lua#L232

bturkus:
not to make this more complicated for me (or anyone), but I'd love to be able to see separate phasecopes for ch 1& 2 and ch 3 & 4

dericed:
@retokromer, vrecord does have an mpv dependency

retokromer:
Could we not use «only» mpv? It is much more versatile than ffplay. Just asking, because nowadays I almost never use ffplay.

privatezero:
@bturkus, I don't think that they were ever successful in integrating the full audio passthrough mode into qc mode, so I think that line is more of a place holder? Current audio passthrough is at https://github.com/amiaopensource/vrecord/blob/master/vrecord#L648

I have some wiggle room this week with my 'research' time, so I'll try to take a peak at this later today to at least get reacquainted with the issues. (Assuming I can finish rebuilding my system from a total meltdown this week hahaha)

privatezero:
@retokromer, it has been a while since I was doing side to side comparisons of ffplay vs mpv in the vrecord context, but at the time (a couple years ago) they definitely each had their own quirks that made them better for some things and worse for others in vrecord - but sadly my memory has faded on what exactly worked better in ffplay. Might not be a difference any more!

[...]

PR #437, opened 2019-05-31
start changing around filters

privatezero:
Changes to audio passthrough mode (`vrecord -a`) to center it more around audio information. Added phase scope for channels 1/2, phase scope for channels 3/4 (hopefully partially addressing https://github.com/amiaopensource/vrecord/issues/434) , audio wave and audio spectrum filters.

Don't have a VCR at my desk rn, so tested this with virtual file input/an empty Blackmagic signal, so please test and let me know if sizing etc looks on other people's end, and if there are any weird issues! (As well as if it actually works correctly on a four channel audio input).

audiopassthrough

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -646,16 +646,14 @@ _passthrough_mode(){
 
 _audiopassthrough_mode(){
     PLAYBACKFILTER="\
-[aid1]asplit=2[z][ao],\
-[z]channelsplit=channel_layout=quad[s1][s2][s3][s4];[s1][s2][s3][s4]amerge=inputs=4,aformat=channel_layouts=quad[zz],\
-[zz]showvolume=t=0:h=17:w=200[xx],\
-[vid1]split=5[a][b][c][d][e],\
-[b]field=top,${WAVEFORM_FILTER}[b1],\
-[c]field=bottom,${WAVEFORM_FILTER}[c1],\
-[d]${VECTORSCOPE_FILTER}[d1],\
-[e]scale=512:ih,signalstats=out=brng[e1],\
-[a][b1][c1][e1][d1]xstack=inputs=5:layout=0_0|0_h0|0_h0+h1|w0_0|w0_h0[abcde1],\
-[abcde1][xx]overlay=10:10[vo]"
+[aid1]asplit=6[a][b][c][d][e][ao],\
+[a]showvolume=t=0:h=17:w=200[a1],\
+[b]pan=stereo|c0=c0|c1=c1,avectorscope[b1],\
+[c]pan=stereo|c0=c2|c1=c3,avectorscope[c1],\
+[d]showspectrum=s=535x672:color=rainbow:legend=1[d1],\
+[e]showwaves=split_channels=1:s=500x500:mode=cline[e1],\
+[vid1][b1][c1][e1][d1]xstack=inputs=5:layout=0_0|0_h0|0_h0+h1|w0_0|w1_h0[abcde1],\
+[abcde1][a1]overlay=10:10[vo]"
     echo "ESC quit" > ~/.config/mpv/input.conf
     _check_mpv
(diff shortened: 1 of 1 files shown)


bturkus:
I went to try this out, but for some reason I couldn't pull up the new view. It looks the key binding isn't connected or something, but I may be testing incorrectly. Thanks for working on this!

privatezero:
Hey @bturkus! The key binding isn't set up for the qc view mode yet - this just works with the old stlye audio passthrough mode which can be fired up with `vrecord -p`. I figured integrate it the easy way first and then if it worked (and the filters are what people want) I will look at bringing it over into qc view!

privatezero:
hahaha and by `vrecord -p` I mean `vrecord -a` durrrrr

bturkus:
I coulda sworn I tried that. Either way, I got it working and this is soooo cool! But it looks to me like I'm only getting CH1&2 (for 4 CH VHS w/ Hi-Fi). I believe that I'm wired correctly, and I made sure to change the config menu to 2 stereo tracks. Any ideas? And yes: this is Lord of the Dance!

Screen Shot 2019-06-03 at 12 46 22 PM

kieranjol:
Lol lord of the dance

kieranjol:
My email replies don't work anymore so here's my original message :
LOL lord of the dance

privatezero:
I'm sure you are wired correctly! I just probably didn't get the filter set up right to grab channels 3/4 (didn't have a test source so I was guessing). I'll take another poke at it! More Michael Flatley!!!

privatezero:
also - @kieranjol, looks to me like your email is coming through still!

kieranjol:
Aye,soz for the derail but +-1 LOTD. Andrew you're awesome and your
contributions are 💯

privatezero:
Hi @bturkus - I just tested with a virtual four-channel input and it was working - out of curiosity, does capturing in the dual-stereo mode result in dual-stereo files on your end?

The 'audiopassthrough' filter is currently independent of those settings, so it would be interesting to see if one is working while the other is failing. Any differences in behavior might help pin down what is going awry! The guts of the filter were written pre-switch to the FFmpeg decklink device, so I wonder if anything has changed in how it handles the incoming signal vs bmdtools

image

bturkus:
Hi @privatezero I just tested again with a different set up (4 CH Betacam, straight into BM Ultrastudio HD Mini), and unfortunately, I'm getting the same results. I also did a few dual-stereo test captures (I figured that the config settings were independent of audio passthrough, but I wanted to eliminate any potential issues), and as far as I can tell they're resulting in dual-stereo files. It gets a little tricky when the content is the same across all four channels, but I turned the levels down on 3&4 and the file seemed to reflect the adjustment accordingly.

privatezero:
@bturkus thanks!! That's great to hear! I'll go through the chain to figure out where the two channels are getting mixed out when captured via the actual instead of virtual input chain.

dericed:
I think the commit names are misleading if the format for piping is changed.

privatezero:
@dericed - reverted that unintended merge from testing

privatezero:
I am thinking of tweaking this PR to only monitor the first two audio inputs for the time being, so that it can be merged and not fall too far out of sync with the main branch. (Then I could dive a bit deeper into making it work for a quad track capture in a future PR). This would only partially address @bturkus's issue, but would be a first step!

Relatedly - I was experimenting with integrating the audio filters into the qcview mode and was successful, but there seemed to be an increase in latency when shifting the lua script over to using `lavfi-complex` as an input rather than `vf lavfi=[ ]`. I'll keep testing when I have some time though!

retokromer:
>I am thinking of tweaking this PR to only monitor the first two audio inputs for the time being, so that it can be merged and not fall too far out of sync with the main branch.

I agree.

PR #441, opened 2019-06-15
simplify vtest man

retokromer:
- delete not needed escape
- simplify commands


vtest.1
```diff
@@ -1,35 +1,35 @@
-.TH vtest 1 "https://github.com/amiaopensource/vrecord" "2018\-06\-07" "AMIA Open Source"
+.TH vtest 1 "https://github.com/amiaopensource/vrecord" "2019-05-03" "AMIA Open Source"
```
privatezero: I'm about to do a new release to reflect the latest merged bug-fix, so you might want to change this to 2019-06-17 to reflect that! I'll hold of on the release for a bit if you would like to get this PR integrated into it.

retokromer: >I'll hold of on the release for a bit if you would like to get this PR integrated into it.

That’s very fine with me.

PR #442, opened 2019-06-24
Further changes for linux/docs

privatezero:
* Fixes problem with `gzcat` vs `zcat`
* Starts more detailed instructions for full install on Linux

### DIFF
--- a/Resources/Documentation/linux_installation.md
+++ b/Resources/Documentation/linux_installation.md
@@ -1,0 +1,44 @@
+# Steps for a successful install of Vrecord on Linux (Ubuntu 18.04)
+
+## About
+There are many possible ways to install the various dependencies of Vrecord on Linux. The following instructions aim to minimize use of linuxbrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord (tested on Ubuntu 18.04)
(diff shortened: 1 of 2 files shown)


retokromer:
@privatezero Do you really need to install dependencies with `apt-get` on your end? 

(We use since two years or so (version `0.7.28`) Debian in production und Ubuntu a little in production but more for teaching.)

privatezero:
@retokromer, yes, I have had rotten luck with having linuxbrew complete full/sustainable installs on my end, usually due to tangled dependencies, which has led me to favor `apt-get` when possible. Are you saying you have had no problems with a full install using only the vrecord brew formula?

retokromer:
@privatezero Our installation script sets up working stations with various software elements we use for AV preservation, including vrecord. I will be back at the lab by end of next week and can check the order of the installations and possible dependencies between them. In general, today we have less difficulties with Linux (Debian) than with macOS, that’s for sure.

retokromer:
@privatezero I just run a test and I confirm that on my end I can install vrecord on Debian using only brew.

privatezero:
Thanks for the follow-up @retokromer, I'll have to spin up a Debian machine and see if it works better than Ubuntu for me!

PR #444, opened 2019-06-29
Conditional gui

dericed:
Most refactoring how selection lists are presented in the gui, then added some new conditional features.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -220,11 +220,11 @@ _gtk_vbox_list() {
 
     _get_list_extras(){
         if [[ "${VARIABLE_NAME}" == "VIDEO_CODEC_CHOICE" ]] ; then
-            echo "<action condition=\"command_is_true( [[ \$VIDEO_CODEC_CHOICE == 'FFV1 version 3' ]] && echo true)\">enable:FFV1_SLICE_CHOICE</action>
-                  <action condition=\"command_is_true( [[ \$VIDEO_CODEC_CHOICE != 'FFV1 version 3' ]] && echo true)\">disable:FFV1_SLICE_CHOICE</action>"
+            echo '<action condition="command_is_true( [ \"$VIDEO_CODEC_CHOICE\" = \"FFV1 version 3\" ] && echo true)">enable:FFV1_SLICE_CHOICE</action>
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -210,6 +210,47 @@ _get_decklink_inputs(){
...
+    local -n OPTION_LIST="${2}"
+    LABEL="${3}"
+    SELECTION="$(_get_index_of_value "${!VARIABLE_NAME}" "${OPTION_LIST[@]}")"
+    LIST="$(_expand_list2items "${OPTION_LIST[@]}")"
+
+    _get_list_extras(){
+        if [[ "${VARIABLE_NAME}" == "VIDEO_CODEC_CHOICE" ]] ; then
+            echo "<action condition=\"command_is_true( [[ \$VIDEO_CODEC_CHOICE == 'FFV1 version 3' ]] && echo true)\">enable:FFV1_SLICE_CHOICE</action>
+                  <action condition=\"command_is_true( [[ \$VIDEO_CODEC_CHOICE != 'FFV1 version 3' ]] && echo true)\">disable:FFV1_SLICE_CHOICE</action>"
+        elif [[ "${VARIABLE_NAME}" == "CONTAINER_CHOICE" ]] ;then
```
retokromer: nit: add a space between `;then`

privatezero:
Appears to have some issues on linux due to it dropping into shell while loading gtk-dialog and not being able to use `[[ ]]` - fixing this half fixes the problem, but needs some more poking to get the escaping/quoting correct

retokromer:
Confirming issues on Debian.

vrecord
```diff
@@ -210,6 +210,47 @@ _get_decklink_inputs(){
     fi
 }
 
+_gtk_vbox_list() {
+    VARIABLE_NAME="${1}"
+    local -n OPTION_LIST="${2}"
```
dericed: Not merging yet, since this use of `local -n` or `declare -n` requires bash 4.3+ which isn't a stated dependency

dericed:
This patch is updated (specifically the `move gui vbox list options to a function` commit is updated) and now works on a Mac with bash 3.2. I still see an issue that the disable commands don't work on startup. For instance if ProRes was the last selected codec, then when running `vrecord -e` the ffv1-slice box is enabled, then changing the codec to a non-FFv1 option disables it.

dericed:
thanks @ablwr @privatezero

PR #445, opened 2019-07-14
Escape XML/Shell with Linux/sh support

ablwr:
I think that while the default shell on Mac is Bash, Linux follows a more simple and strict shell pattern, dash, which doesn't support several bash features.

Or maybe things are just wacky!

Let me know if this also works on a macOS.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -219,11 +219,11 @@ _gtk_vbox_list() {
 
     _get_list_extras(){
         if [[ "${VARIABLE_NAME}" == "VIDEO_CODEC_CHOICE" ]] ; then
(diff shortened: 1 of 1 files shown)


ablwr:
I read about the differences in support here: http://mywiki.wooledge.org/Bashism

retokromer:
>I think that while the default shell on Mac is Bash,

That has changed in macOS 10.15 (Death Valley) where `zsh` is the standard shell.

>Linux follows a more simple and strict shell pattern,

Bash works well on Linux! We use it daily since years.

dericed:
Closing this as this patch is approved but based on an unapproved pr, but this patch has been cherry-picked into that pr at https://github.com/amiaopensource/vrecord/pull/444/commits. Thanks @ablwr.

PR #454, opened 2019-08-02
add teranex to hardware

bturkus:


retokromer:
Did you test is with Express and/or AV?

bturkus:
I’ve tested with multiple generations of 2Ds, which were discontinued last year and replaced by (I believe) the AV. Do have reason to suspect that it won’t work with either of the current models?

retokromer:
>Do have reason to suspect that it won’t work with either of the current models?

I was wondering, because we had problems in the past, when Blackmagic changed configurations and settings of the same products without any notice.

PR #455, opened 2019-08-04
use `<kbd>`

retokromer:


dericed:
any votes here? I feel a bit mixed about introducing html into a markdown doc that doesn't otherwise use it.

retokromer:
@dericed Alternatively, we could only unify `ctrl` and `control`.

retokromer:
No votes. Closing.

PR #470, opened 2019-10-03
Update linux_installation.md

akynr:
It seems the option for enabling DV devices  with ffmpegdecklink changed to --with-iec61883

`~$ brew install ffmpegdecklink --help
Usage: brew install [options] formula

[...]
        **--with-iec61883              ffmpegdecklink: Enable DV device (Linux)**
[...]
    -h, --help                       Show this message.`


dericed:
I'd suggest including this first in https://github.com/amiaopensource/homebrew-amiaos/ on the ffmpegdecklink installer as the --with-iec61883 could be enabled there if the installer is running on Linux. Once in there, then we can add an alternate input to vrecord, as I've been drafting for avfoundation in one of the recent branches.

privatezero:
Thanks for the catch @akynr!

Issue #452, opened 2019-08-02
Identifier Not Valid Error


bturkus:
hey

on the latest version of vrecord (2019-07-22), we're seeing an "Error: Not a valid option, select a valid number" message, no matter what we name our identifiers. Is anyone else seeing this?

ben

retokromer:
It’s generated by the `_lookup_choice()` function:
https://github.com/amiaopensource/vrecord/blob/0f6d3ed31e52035625987555bc70c4811d96ce7d/vrecord#L923-L925

dericed:
Hi, @bturkus, it means that a selection you made in unexpected. Can you recreate the issue? If so please send a screenshot of the gui before you start the recording or provide more of the command line output.

bturkus:
here's a screenshot of the GUI pre-error. I cannot figure out where I've gone wrong. Thanks for looking into it!

<img width="1605" alt="Screen Shot 2019-08-05 at 2 30 35 PM" src="https://user-images.githubusercontent.com/8987230/62486587-b23b1900-b78d-11e9-97d2-51e1c4c40378.png">

dericed:
can you run `cat ~/.vrecord.conf ` and send the result (censor if needed).

bturkus:
```
MY-LPAMI-056430:~ benjaminturkus$ cat ~/.vrecord.conf
# Set these variables to a valid option or leave as empty quotes (like "") to request each run.
VIDEO_INPUT_CHOICE="SDI"
AUDIO_INPUT_CHOICE="SDI Embedded Audio"
CONTAINER_CHOICE="Matroska"
VIDEO_CODEC_CHOICE="FFV1 version 3"
FFV1_SLICE_CHOICE="24"
AUDIO_CODEC_CHOICE="24-bit FLAC"
VIDEO_BIT_DEPTH_CHOICE="10 bit"
AUDIO_MAPPING_CHOICE="1 Stereo Track (From Channels 1 & 2)"
TIMECODE_CHOICE="none"
STANDARD_CHOICE="NTSC"
QCTOOLSXML_CHOICE="Yes, after recording"
FRAMEMD5_CHOICE="Yes"
EMBED_LOGS_CHOICE="Yes"
PLAYBACKVIEW_CHOICE="Quality Control View (mpv)"
PLAYBACKVIEW_CHOICE_PASS="Quality Control View (mpv)"
DIR="/Users/benjaminturkus/Desktop/vrecord_tests"
[...]

PR #471, opened 2019-10-13
remove unneeded function call

dericed:
The FFV1_SLICE_CHOICE is simply the integer needed, so there is no need to analyze or lookup that value with the _review_option function.

fixed https://github.com/amiaopensource/vrecord/issues/452

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1093,20 +1093,31 @@ if [[ "${INVERT_PHASE}" = 'true' ]] ; then
 fi
 
 _review_option(){
+    OPTIND=1
+    while getopts "n" opt ; do
+        case "${opt}" in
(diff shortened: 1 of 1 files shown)


dericed:
Do you mean to the config file (so that the selection is remembered between instances of vrecord)? That happens in https://github.com/amiaopensource/vrecord/blob/b37e734cbce5774cca6b9503d777e7f95ffbe36a/vrecord#L535.

privatezero:
Nah, I was looking at the array in https://github.com/amiaopensource/vrecord/blob/b37e734cbce5774cca6b9503d777e7f95ffbe36a/vrecord#L1103 that is part of the `_review_option` function and is ultimately written to the capture log. (Again only have time to skim so lemme know if I am missing other places!!)

dericed:
@privatezero updated to another approach, please review.

Issue #322, opened 2018-05-13
[wish] other graphic cards than Blackmagic’s


retokromer:
I know very well that there are some legal issues, but I suggest nevertheless to support other graphic cards as well, _in primis_ AJA’s, because this is the reality in the field.

dericed:
I suggest that the scope of vrecord is not to develop support for hardware but to build upon what support for hardware is already developed in open projects, ie bmdtools or decklink's support in libavdevice. If you contribute an aja input device to libavdevice, then we could consider integrating it.

PR #473, opened 2019-10-24
adds an avfoundation input option to vrecord

dericed:
supports capture from DV devices. The options are very limited to focus on piping the dv stream from the input to output, so the interface is a lot simpler than that for decklink. Rewrapping to a container could be added as a post-process later.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -577,8 +577,8 @@ _passthrough_mode(){
     _set_ffplay_options
     if [[ "${MEDIA_PLAYER_CHOICE}" = "mpv" ]] ; then
         _check_mpv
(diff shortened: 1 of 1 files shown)


dericed:
documentation is needed for this before merge

dericed:
It's a closed issue but this PR would resolve https://github.com/amiaopensource/vrecord/issues/322 since the device input is not the decklink input.

dericed:
going to merge, @libbyhopfauf will follow with a PR for the documentation before the next release.

libbyhopfauf:
on it!

PR #475, opened 2019-10-29
dv update to documentation

libbyhopfauf:
# DV Capture

To capture DV formats in vrecord, you will need to have your DV deck connected directly to your computer via FireWire input. Make sure your deck is in “Local” mode. At this time, vrecord is not equipped with tape control, so you will need to manually start and stop the tape just like with analog capture in vrecord. 

## DV Settings
dv_vrecord_configuration

1. To select the deck you would like to use, go into the configuration mode either by:
    1. run `vrecord -e`
    1. clicking on the “Edit Settings” button in the GUI (run `vrecord -g`)
4. Switch from the "Decklink" tab to the "DV" tab under “Input Options” at the top of the configuration window.
5. Select the name of the DV device you want to use from the list.
6. Specify the playback, sidecar, file naming, recording event and directory options as you normally would for vrecord. For details see [Editing Settings](Resources/Documentation/settings.md)
7. Click “OK” to save.
8. You can then run passthrough and record modes the same as you would with analog videotape. Please note that if the timecode does not start at the very beginning of the tape, the record mode viewer will not open until the timecode is detected (as soon as the counter starts moving on your deck, the viewer window should pop-up)

## Bitstream Error Concealment
If vrecord detects that the DV device is concealing bitstream errors, this will be noted in the Terminal.
dv_vrecord_bitstream_concealment2


kieranjol:
This looks incredible to me, awesome work everyone - this is pretty stunning!

dericed:
I updated the images such to reduce their size a little, adjusting quoting, and added a caveat that this is currently mac only. Otherwise this looks good. I think it's mergeable as is, could also use a link to this new doc from the readme or one of the other resource docs, just so it's easier to navigate to here.

PR #479, opened 2019-12-02
edit filter

privatezero:
Currently certain preview modes do not represent the actual aspect ratio that is being captured to file. Unfiltered view seems to get it right, but the ffplay views that utilize the longer filter chain don't seem to be passing the parameters that are set via NTSC or PAL choices. (See examples below). This PR applies upstream settings into the filter chain.

Broadcast range visual:
image

Unfiltered:
image

Broadcast range visual (this branch):
image

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -900,7 +900,7 @@ _lookup_choice(){
         "Broadcast Range Visual")
             PLAYBACKFILTER="\
 ${PLAYBACK_FILTER_ADJUSTMENT}split=5[a][b][c][d][e];\
-[a]copy${TIMECODE_OVERLAY}[a1];\
+[a]${RECORDINGFILTER}${TIMECODE_OVERLAY}[a1];\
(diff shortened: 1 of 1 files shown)


dericed:
The selection of this depends on if you prioritize seeing all the pixels or seeing the pixels squeezed into the intended display aspect ratio. Before the patch the presentation would include all 720x486 pixels but after the image is rescaled.

privatezero:
Yeah - I mean, maybe I am just capriciously changing my mind on this, but In terms of the user perspective I am thinking that maybe having a view that previews the same scaling as the file being written might be valuable? But I definitely see the other side as well

privatezero:
But, just kinda thinking out loud with this - happy to close if nobody agrees!

retokromer:
Hmm… both points of view are equally valid, in my opinion! Best would possibly be to offer the possibility to chose, but that would make things more complicated.

PR #480, opened 2019-12-04
update log docs

bturkus:
this is correct, right? vrecord only produces 2 logs: ffmpeg log and capture options log


dericed:
thx @bturkus

PR #488, opened 2020-01-16
Add information about RPM install method

lsde:
Here I propose an addition of another install method for Linux, using RPM package based on a .spec file in https://github.com/NFAcz/vrecord-rpm repository which I've been testing lately.

Ideally, I'd like to see this being part of this upstream repo, but since unification would mean adding some detecting logic, this seems to be a simpler method for the time being..

Question: Would you agree to submit the package for inclusion into Fedora and EPEL repos in the future?

(We're not using Debian at all, therefore I didn't manage to create debian packages, but that should be fairly simple in the future..)

### DIFF
--- a/Resources/Documentation/linux_installation.md
+++ b/Resources/Documentation/linux_installation.md
@@ -1,21 +1,24 @@
-# Steps for a successful install of Vrecord on Linux (Ubuntu 18.04)
+# Steps for a successful install of Vrecord on Linux
 
 ## About
-There are many possible ways to install the various dependencies of Vrecord on Linux. The following instructions aim to minimize use of linuxbrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord (tested on Ubuntu 18.04)
+There are many possible ways to install the various dependencies of Vrecord on Linux. 
 
-## Programs to be installed manually
+## via Linuxbrew (tested Ubuntu 18.04)
+The following instructions aim to minimize use of linuxbrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord (tested on Ubuntu 18.04)
+
+### Programs to be installed manually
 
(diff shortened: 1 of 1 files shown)


dericed:
I’m very interested in making this more accessible and expanding use beyond brew environments. Partly an issue is getting ffmpeg build with decklink as it requires nonfree and I’m skeptical that homebrew would welcome building with decklink. Still we could do some detecting to see if the local ffmpeg is built to support it and if so skip any use of brew variables.

retokromer:
>(We're not using Debian at all, therefore I didn't manage to create debian packages, but that should be fairly simple in the future..)

Debian is not difficult.

>I’m skeptical that homebrew would welcome building with decklink.

It has been refused by the maintainers.

retokromer:
>Question: Would you agree to submit the package for inclusion into Fedora and EPEL repos in the future?

+1

Resources/Documentation/linux_installation.md
```diff
@@ -1,21 +1,24 @@
-# Steps for a successful install of Vrecord on Linux (Ubuntu 18.04)
+# Steps for a successful install of Vrecord on Linux
 
 ## About
-There are many possible ways to install the various dependencies of Vrecord on Linux. The following instructions aim to minimize use of linuxbrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord (tested on Ubuntu 18.04)
+There are many possible ways to install the various dependencies of Vrecord on Linux. 
 
-## Programs to be installed manually
+## via Linuxbrew (tested Ubuntu 18.04)
```
retokromer: I guess `###` would be better.

dericed:
I'd like to avoid vrecord-bypass-homebrew.patch. Perhaps we could just test for the presence of `"$(brew --prefix ffmpegdecklink)/bin/ffmpeg-dl"` and if not there then fallback to `ffmpeg`, then have a fallback if the result of `ffmpeg -sources decklink` is empty

dericed:
I filed an issue at https://github.com/amiaopensource/vrecord/issues/500 to deal with helping fallback to the locally installed ffmpeg if the brew ffmpegdecklink is not installed. Merging now, thank up @lsde!

PR #489, opened 2020-01-16
More generic, logging, and input selection

dericed:


privatezero:
On my end, as written this seems to necessitate a user to run `vrecord -e` prior to every session as the `AVF_INPUT_CHOICE` variable is not persisting. This means on my end if I run `vrecord -e` and immediately follow prompts to either record or enter passthrough then vrecord functions as intended, but should I choose to exit and then try capturing it fails with ': Device not configured'

Seems like it would be good to have some kind of fallback default device option?

retokromer:
>Seems like it would be good to have some kind of fallback default device option?

+1

dericed:
> Seems like it would be good to have some kind of fallback default device option?

-1

The avfoundation device list is filled with many options irrelevant to vrecord. Potentially we could iterate each avfoundation option to probe the content to classify what it is. Within decklink it makes sense to default to the first input but here that could be the computer's camera or a screen recording.

I suggest to mandate the input rather than default and added commits to require that value if unset.

retokromer:
>I suggest to mandate the input rather than default and added commits to require that value if unset.

Yep, I guess this is a better solution.

dericed:
Since a default option makes sense for decklink, I moved parts and added a default select if there is only one decklink device attached. Please re-review.

privatezero:
Still having some issues here - 
I get a `: Device not configured` error when attempting to load vrecord not via the `vrecord -e` follow-up prompts.

When the Blackmagic device is unplugged and replugged in on first run vrecord seems to find it but is still unable to use it, with the error `UltraStudio Express: Device not configured`. After running once it reverts to the previous error.

dericed:
@privatezero what os are you using?

privatezero:
Testing so far on mac - should be 10.13.6

dericed:
Thx @privatezero, please review after this last patch

privatezero:
retested and performs well on my end!

PR #492, opened 2020-01-22
update filter/switch to ffplay

privatezero:
Making a new PR for an initial revamping of the audio monitoring (as well as some streamlining of dependencies) - this seems to be stable on my end. Testing appreciated!


privatezero:
@bturkus - I will investigate combining with more video tools in the future!

alavigne12:
Just tested this at MIPoPS and it worked fine!

Issue #474, opened 2019-10-28
vrecord settings / Desktop Video Setup mismatch causes dropped frames warning..


asfinn:
vrecord video and audio settings seem to override the BM Desktop Video Setup input settings but a mismatch does seem to cause a consistent Dropped Frames warning in capture_options.log like this:

DROPPED_FRAMES_TIMESTAMPS: 00:00:00.040 00:00:00.280

I don't see this a a major problem but it's disconcerting and not immediately obvious where the cause lies. We regularly swap between SDI / embedded and analogue / XLR inputs on one of our systems.

dericed:
Hi @asfinn, that's really interesting. I've seen the instances where there would be a couple dropped frames right at the beginning, but didn't correlate it to a mismatch with the BM Desktop Video Setup. Could you provide more information about your experiments? If the consistency of this settings between the Desktop Video and vrecord is a real issue, then perhaps we need to parse the plist file for Desktop Video to warning about the conflict earlier on.

asfinn:
Hi @dericed - yes, they only seem to be at the start and were consistent with configuration mismatch.. I haven't noticed this before but we've only just upgraded to MacOS 10.13.6 and Desktop Video 11.4 and I think this is when it's been more apparent.. Below are the capture_options.log and ffmpeg_decklink_input.log when the DTV and vrecord settings were mismatched:

`computer_name: rebbie.local
computer_model_name: Mac Pro
computer_model_id: MacPro5,1
computer_processor_name: 6-Core Intel Xeon
computer_processor_speed: 3.2 GHz
computer_processor_count: 1
computer_memory: 8 GB
computer_serial: CK92004D4PD
computer_cores: 6
user_name: adrian
operating_system_VERSION: Darwin Kernel Version 17.7.0: Sun Jun  2 20:31:42 PDT 2019; root:xnu-4570.71.46~1/RELEASE_X86_64
vrecord version: 2019-07-22_1
datetime_start: 2019-10-28T14:33:08
capture_device_node: /dev/disk5
capture_device_protocol: SATA
capture_device_name: TOSHIBA HDWE140
capture_device_block_size: 512 Bytes
capture_device_location: Internal
"Bay 1"
capture_device_partition_type: Apple_HFS
capture_device_partition_name: Mac OS Extended (Journaled)
capture_device_volume_uuid: 169B75DD-C409-3ED7-AFFB-96F52997A9A0
capture_device_volume_available: 
[...]

PR #501, opened 2020-02-10
recommend matching input settings between vrecord and desktop video

dericed:
partially addresses https://github.com/amiaopensource/vrecord/issues/474


dericed:
psst @asfinn

PR #505, opened 2020-02-10
mpv fix

bturkus:
think something in latest version of mpv changed, and now an equals sign is needed to get the lua working


privatezero:
Just tested in both new and prior version of mpv - verified the `=` is backwards compatible so don't have to worry about checking for mpv version. yay!

bturkus:
bt view lives!!!!

privatezero:
Oh crap - that reminds me that I gotta do a PR to update documentation to incorporate BT view!

PR #510, opened 2020-02-22
add HuffYUV encoding

### LINKED ISSUE #366: [wish] HuffYUV as capture format
I suggest to add the HuffYUV encoding – of course, in FFmpeg owns flavour `ffvhuff`, not the original `huffyuv` – as a choice. Both formats hold the same «raw» image content – this can be proven e.g. by `framemd5` – and have the same lossless compression rate, but the encoding of HuffYUV does require almost <sup>3</sup>/<sub>4</sub> of the time needed by FFV1. (Speed is the reason why we still use this codec for capturing on low-tech hi-quality film scanners.)

retokromer:
should resolve #366

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -847,7 +847,11 @@ _lookup_choice(){
             SUFFIX="" ;;
         "HuffYUV")
             VIDEOCODECNAME="HuffYUV"
-            MIDDLEOPTIONS+=(-c:v ffvhuff -pix_fmt yuv422p10le -pred median -context 1)
+            if [[ "${PIXEL_FORMAT}" = "yuv422p10" ]] ; then
+                MIDDLEOPTIONS+=(-c:v ffvhuff -pix_fmt yuv422p10le -pred median -context 1)
+            elif [[ "${PIXEL_FORMAT}" = "uyvy422" ]] ; then
+                MIDDLEOPTIONS+=(-c:v ffvhuff -pix_fmt yuv422 -pred median -context 1)
+            fi
             SUFFIX="_huff" ;;
 
         # video pixel format and bit depth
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -845,6 +845,11 @@ _lookup_choice(){
             VIDEOCODECNAME="H.264"
             MIDDLEOPTIONS+=(-c:v libx264 -crf 18 -pix_fmt yuv420p)
             SUFFIX="" ;;
+        "HuffYUV")
+            VIDEOCODECNAME="HuffYUV"
+            MIDDLEOPTIONS+=(-c:v ffvhuff -pix_fmt yuv422p10le -pred median -context 1)
```
dericed: Could you add a bit depth condition similar to what is used in Uncompressed Video.

retokromer: You mean the possibility of choose between 8 and 10 bit?

dericed: Yes, as written, the user may select 8 bit as a decklink input option, but the huffyuv would be forced to 8 bit. I suggest the pix_fmt of the huffyuv encoding match the input (whether the decklink is set to 8 or 10), the same logic is used in the Uncompressed Video encoding.

retokromer: I’ll amend the PR that way.

retokromer: HuffYUV does not support `uyvy422`, I will use `yuv422` instead.

dericed: `Unknown pixel format requested: yuv422.`

retokromer: Oops, sorry! Planar missing.

retokromer:
Is there anything else that needs to be done, approved or reviewed before squashing and merging this?

privatezero:
Code looks good to me, I haven't tested this yet - I can give it a run next week

retokromer:
Thank you, @privatezero! Once merged and a new release made, I can delete my alternate formula.

privatezero:
Just tested 8 bit and 10 bit capture - seems to work as intended

PR #511, opened 2020-02-23
make doc easier to maintain

retokromer:
You don’t need to renumber the other items when adding or deleting one.

### DIFF
--- a/Resources/Documentation/dv_info.md
+++ b/Resources/Documentation/dv_info.md
@@ -6,16 +6,16 @@ To capture DV formats in vrecord (when installed in macOS), you will need to hav
 ![Alt text](../dv_vrecord_configuration.png "DV Capture")
 
 1. To select the deck you would like to use, go into the configuration mode either by:
-- run `vrecord -e`
-- clicking on the "Edit Settings" button in the GUI (run `vrecord -g`)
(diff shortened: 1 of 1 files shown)


dericed:
I'm unsure. Maybe easier for some to maintain, but incrementing integers is well-known, and it's more confusing to read if unrendered.  @privatezero, any opinion?

privatezero:
Yeah, I can see pluses and minuses. Just checked and the basic pandoc md to html conversion renders correctly with this edit, so if that route is taken in the future for local renders this doesn't affect things. Overall, I am OK with this change.

retokromer:
Wider context FYI: The content was dissociated from the layout in HTML 3.2 (second half of the 1990s, if I remember carefully), introducing XML and CSS. We did it in order to simplify the work. [However, we didn’t imagine back then how relevant XML would become. The majority of us just considered SGML too complex and tried another route.]

With this change one can add, move or delete items from an ordered list without having to take care of the correct numbering.

@privatezero: Thanks for checking! I was quite sure `pandoc` would digest it correctly, but I didn’t test it…

retokromer:
For reference: https://guides.github.com/features/mastering-markdown/

privatezero:
Roger that, @retokromer

PR #529, opened 2020-04-19
optional dependencies

dericed:


privatezero:
I did notice that the `refresh` button in GUI does not reflect dependency changes - maybe not much of a priority though

PR #530, opened 2020-04-21
set avfoundation input to not drop late frames

dericed:


retokromer:
To my knowledge it’s not in the current release: https://github.com/FFmpeg/FFmpeg/commit/2ca14d84eefd3178822d59347f8978d93230f54c

retokromer:
@dericed : I tried to answer @privatezero’s question.

dericed:
ah sorry, @privatezero and I discussed via another means in sync. The flag is patched into ffmpegdecklink for now at https://github.com/amiaopensource/homebrew-amiaos/blob/master/ffmpegdecklink.rb#L112.

PR #543, opened 2020-05-27
Audio Mode

privatezero:
This probably should have a notice added to the GUI that it is still 'beta' and needs some instructions for external device control on macOS.

* Adds initial support of audio capture through ALSA (linux) or avfoundation (macOS)
* Separate views for passthrough and capture of audio to try for safer captures
* FLAC and PCM support
* macOS needs to control input device settings (sample rate and bit depth) through external midi control utility
* avfoundation seems laggy at 96 kHz during passthrough filter, works fine with ALSA


bturkus:
Thanks for working on this @privatezero!! This is gonna be huge for @XFRCollective and many others who are looking for stream-lined, cost-effective solutions.

privatezero:
Thanks @bturkus  - hopefully will have it a little more polished soon!

PR #552, opened 2020-06-12
add in avfctl

dericed:
This is a work in progress as part of this broken the _get_output_filename function. The avfctl testing is mac only but this impacts the conditions around passthrough and record for all tabs so decklink and audio playback and recording should be closely reviewed.


dericed:
I fixed the output filename issue and rebased. Ready for review.

retokromer:
Sorry, I cannot test at the moment, because of holidays.

Issue #566, opened 2020-07-15
audio capture mode issues


bturkus:
Apologies for the not-very-descriptive issue title, but I'm back at work testing out new vrecord (so much cool stuff!!), and I'm encountering a few issues with audio capture mode. I'm pretty sure it's user error, but any help in determining where I'm going wrong would be majorly appreciated.

So I've got an audiocassette deck, analogs out going straight in to my Ultrastudio HD Mini (analog ins), and I've disconnected all video-related cables. I changed the Sound preferences in System Prefs (macOS), and I changed Desktop Video setup to analog audio in, but when I go to capture as audio in vrecord, I'm not getting any signal. In Media Express, I do see the audio coming in, so I'm not sure exactly what I'm missing (I also made sure to change the audio input in the Decklink options in vrecord proper). 

Additionally, when I select passthrough mode from the audio capture tab, I get a ffmpeg-dl error: 
```
    Stream #0:0: Video: rawvideo (Y4[11][8] / 0x80B3459), yuva420p, 1439x728 [SAR 1:1 DAR 1439:728], 25 tbr, 25 tbn, 25 tbc
    Stream #0:1: Audio: pcm_s16le, 48000 Hz, stereo, s16, 1536 kb/s
/usr/local/bin/vrecord: line 1215: 72483 Exit 1                  "${FFMPEG_BIN}" -nostdin -hide_banner -nostats "${GRAB_INPUT[@]}" "${PIPE_OUTPUT[@]}" 2> "${VRECORD_INPUT_TMP}"
     72484 Segmentation fault: 11  | "${FFPLAY_BIN}" -window_title "${WINDOW_NAME}" -f lavfi "amovie='pipe\:0',${PLAYBACKFILTER}"
```
Is this to be expected, as in, passthrough is not an option for audio capture mode?

I'm assuming @privatezero will be the one responding to this, and just wanna say again: thanks for making this happen!!

Ben

privatezero:
Hi @bturkus, unfortunately we are only halfway there for what you need in audio mode! Right now, it doesn't seem to work with Blackmagic devices - It only works with avfoundation inputs (or alsa on Linux). Weirdly, Blackmagic devices will show up in the scan of avfoundation devices, but they won't actually do a successful capture.

[...]

PR #569, opened 2020-07-16
force overlay to yuv444 in this case

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/566 with ffmpeg 4.3 and 4.3.1


privatezero:
this is working on my end as well!

retokromer:
I suggest to make a new release, because this is a bug fix.

privatezero:
Agree, but this should be merged first - https://github.com/amiaopensource/vrecord/pull/570

retokromer:
Yep.

PR #571, opened 2020-07-19
Caption viewer

dericed:


bturkus:
testing this now...I think there's a small gtkdialog design issue. This is where you set the line for the caption reader, right?

Screen Shot 2020-07-20 at 10 33 33 AM

dericed:
ya, guess it needs a label. Also maybe it makes more sense as a very small range of integers rather than a free text box.

Issue #574, opened 2020-07-21
UI Request: It's currently difficult to find external media when setting recording directory


iamdamosuzuki:
I was just helping out a colleague who couldn't figure out how to find their external hardware in the window that opens when you want to select the recording directory. Currently, you need to go to click "File Systems" on the left side, then "Volumes" to see your external media. This may be obvious for someone familiar with Unix paths, but I can see how this can be confusing. Is there a way to fix the directory selection window, like having it show external media on the left side column? Or is this something that should be addressed in the documentation? I'm happy to update the documentation if that's the best  route to take.

PR #576, opened 2020-07-31
add some guidance to directory selector

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/574

@iamdamosuzuki, can you review this?

The result looks like this:
image


dericed:
updated. Thx @privatezero for the review.

PR #579, opened 2020-08-05
Linux dv work

privatezero:
Doesn't currently allow device selection (possibly relating to https://trac.ffmpeg.org/ticket/5966) but establishes initial support of DV preview and capture on Linux

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -2341,19 +2341,22 @@ if [[ "${DEVICE_INPUT_CHOICE}" = 0 ]] ; then
     fi
 elif [[ "${DEVICE_INPUT_CHOICE}" = 1 ]] ; then
     if [[ "${OS_TYPE}" = "linux" ]] ; then
-      VRECORD_STEPS="2"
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -544,6 +544,9 @@ _get_avfctl_inputs(){
     if [[ "${#AVFCTL_DEVICES[@]}" = 1 ]] ; then # default to first input if only one
       AVFCTL_INPUT_CHOICE="${AVFCTL_DEVICES[0]}"
     fi
+    if [[ "${OS_TYPE}" = "linux" ]] ; then
+      AVFCTL_DEVICES+=('Default DV Device')
+    fi
```
dericed, line 549: This is just to fill the dropdown with something, right?

privatezero: yup - to fill the drop and let the user know that they have no choice except default. Also, it avoids an error being thrown when AVFCTL_INPUT_CHOICE isn't set, and populates the log. A future PR could probably do a better job of parsing out the name of the default device, but with the behavior experienced in that ffmpeg issue, I didn't want to risk giving false information to the user.

PR #580, opened 2020-08-05
update for release 2020-08-05

privatezero:


retokromer:
Does the `README.md` correspond to this version? If so, then I would suggest to update also this date; otherwise I would suggest to update the documentation as well.

Issue #588, opened 2020-09-03
Line 390 Syntax Error


alexhabgood:
Hello,

I've recently updated to the latest build of vrecord (2020-08-05). However, I now get the following error whenever I try to start it up:

/usr/local/bin/vrecord: line 390: syntax error near unexpected token `;;'
/usr/local/bin/vrecord: line 390: `            "FFV1 version 3")   SUFFIX="_ffv1" ;;'

I've tried using the -e & -x to reset and edit the configuration file, but get stopped with the same error. 

It could be an issue with my set up? It's running on a Mac with High Sierra 10.13.6.

Thanks,

dericed:
could you run `brew install bash` and retry

alexhabgood:
brew install bash gives me:

bash 5.0.18 is already installed and up-to-date
To reinstall 5.0.18, run `brew reinstall bash`

So I reinstalled bash and restarted the terminal. But I'm still getting the same error with vrecord.

alexhabgood:
brew install bash gives me:

bash 5.0.18 is already installed and up-to-date
To reinstall 5.0.18, run `brew reinstall bash`

So I reinstalled bash and restarted the terminal. But I'm still getting the same error with vrecord.

dericed:
thx, i have a fix in mind and can try to post it tonight

alexhabgood:
Great! Thanks for helping me out.

retokromer:
I guess, it’s running the old Bash 3.2.57, kindly provided by Apple, rather than the current one, installed by Homebrew.

alexhabgood:
I initially thought this was the problem too, it having an issue with case. So, I added the updated shell version to /etc/shells and used chsh -s to set it as the default. Now when I echo $BASH_VERSION it shows 5.0.18. Did I miss a step?

retokromer:
You can check with `bash -c 'echo $BASH_VERSION'` which version is actually running, when multiple are installed.

alexhabgood:
When I enter bash -c 'echo $BASH_VERSION', it comes up with the 3.2.57 version. So looks like this could be the problem!

I'll see if I can get this properly switched over tommorow.

retokromer:
Yep, that seems to be the problem. Possibly, this may be useful: Daniel Weibel: [Upgrading Bash on macOS](https://itnext.io/upgrading-bash-on-macos-7138bd1066ba). But, of course, you can also wait for Dave’s patch.

[...]

PR #589, opened 2020-09-03
Separate functions file

dericed:
possible fix for https://github.com/amiaopensource/vrecord/issues/588

note that Resources/vrecord_functions would need to be added to the installer (and installer could possibly remove requirement for brew:bash)


privatezero (approved):
This doesn't seem to break anything on my end - but can't test if it fixes initial problem or not!

PR #599, opened 2020-09-30
prevent recursive logging of variables

dericed:
These would just be continuously adding to the variable file at each launch


privatezero (approved):
lol lgtm!

PR #608, opened 2020-10-13
Add waveform scale option

dericed:


vrecord
```diff
@@ -1931,6 +1940,7 @@ fi
 while [[ "${RUNTYPE}" = "edit" || "${RUNTYPE}" = "GUI" ]] ; do
     _edit_mode
 done
+WAVEFORM_SCALE_OPTIONS=("digital" "millivolts" "ire")
```
privatezero, line 1943: needs to be moved to line 1926 to work I think

dericed: updated

privatezero (approved):
lgtm!

Issue #516, opened 2020-03-01
ProRes: add test that not 8-bit input is not chosen


retokromer:
Currently the user can select 8 bit as a DeckLink input option, but ProRes 422 is a 10-bit format. I suggest to add a test.

retokromer:
Thanks!

PR #610, opened 2020-10-16
add warning if 10 bit prores is used with an 8 bit input

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/516

### DIFF
--- a/Resources/vrecord_functions
+++ b/Resources/vrecord_functions
@@ -85,6 +85,9 @@ _validate_form(){
     elif [ "${AUDIO_CODEC_CHOICE}" = "24-bit FLAC" ] && [ "${CONTAINER_CHOICE}" = "QuickTime" ] ; then
         echo "ERROR: Incompatible audio codecs and containers have been selected.\n"
     fi
+    if [ "${VIDEO_CODEC_CHOICE}" = "ProRes" -o "${VIDEO_CODEC_CHOICE}" = "ProRes (HQ)" ] && [ "${VIDEO_BIT_DEPTH_CHOICE}" = "8 bit" ] ; then
+        echo "WARNING: ProRes (which is a 10 bit encoding) is selected but the input device is set to use 8 bits.\n"
+    fi
     if [ ! -d "${DIR}" ] ; then
         echo "ERROR: The recording directory (${DIR}) does not exist. Create this folder please.\n"
     fi


Resources/vrecord_functions
```diff
@@ -85,6 +85,9 @@ _validate_form(){
     elif [ "${AUDIO_CODEC_CHOICE}" = "24-bit FLAC" ] && [ "${CONTAINER_CHOICE}" = "QuickTime" ] ; then
         echo "ERROR: Incompatible audio codecs and containers have been selected.\n"
     fi
+    if [ "${VIDEO_CODEC_CHOICE}" = "ProRes" -o "${VIDEO_CODEC_CHOICE}" = "ProRes (HQ)" ] && [ "${VIDEO_BIT_DEPTH_CHOICE}" = "8 bit" ] ; then
```
retokromer, line 88: `[[ "${VIDEO_CODEC_CHOICE}" =~ "ProRes" ]]` would trap both.

retokromer (approved):
LGTM (just a possible shorter code in my note)

Resources/vrecord_functions
dericed: I'm cautious to use that formatting in the shell that is supported by gtkdialog. If you've tested and it does, feel welcome to send a patch.

retokromer: I understand. I guess it’s Bash 3 compatible, therefore I would expect that it works in gtkdialog, but I haven’t tested so far.

dericed: `[[ "${VIDEO_CODEC_CHOICE}" =~ "ProRes" ]]` works here. @privatezero, can you test in your environment, iirc we had reasons to avoid `[[`.

retokromer: I ran a test: it works here on Debian (on Qubes). I guess @privatezero is on the more orthodox Ubuntu.

privatezero: Can confirm that change is no bueno on my end - GTK Dialog drops down into sh, not bash for me

privatezero (approved):
Seems to work!

Resources/vrecord_functions
retokromer: Are you on Ubuntu? If so, which one?

privatezero: yup -  still running 18.04, but will probably upgrade to 20.04 soon

dericed: I'm going to consider this issue not blocking, since the issue is with a theoretical proposal and not this PR. Let's move this into another pull request or issue if important enough to deserve it.

retokromer: No, it’s not blocking at all on my end. I’m just curious! I presume it’s the default configuration on Ubuntu, but I’ll check on 20.04.1 LTS, which I should also have on Qubes.

retokromer: FYI @dericed @privatezero : It works fine as well on the Ubuntu stations we have in production. It’s indeed a configuration issue, but I thinks that it would be an overkill to change this parameter either temporally at script’s start or definitively at script’s installation.

PR #620, opened 2020-10-17
Reordering

dericed:
This PR should impact any use, but it groups the functions together at the top, so here lines 15-1849 are functions and from 1851-2407 (the end) are not. I also removed an unused function and fixed an issue where vrecord and vrecord_function shared a function name for two different (though related) functions. I'm hoping this makes it easier to cleanup things. I'm looking particularly at cleaning up the logic in lines 2009-2050 and maybe removing the cli interview stuff in 2052-2084.


retokromer:
Excellent idea to reorder the code!

privatezero (approved):

PR #624, opened 2020-10-21
Fix record if audio+video is set for passthrough

dericed:
fixes https://github.com/amiaopensource/vrecord/issues/622#issue-726704578


dericed:
This PR is on top of https://github.com/amiaopensource/vrecord/pull/620 so I recommend reviewing that one first.

privatezero (approved):

PR #627, opened 2020-10-22
Brng fullrng to config

privatezero:
Shifts choice of `Broadcast Range` or `Full Range` to config tab so that filters can be simplified and Audio + Video can use either setting


dericed (approved):
Hi @privatezero, I did some updates to your PR, but if this works for you, it works for me.

retokromer (approved):

PR #633, opened 2020-10-27
grep instead of wildcards

privatezero:


dericed (approved):
I tested the blackmagic condition on a mac. Worked for me.

Issue #632, opened 2020-10-27
phase/polarity inversion CH3&4


bturkus:
Hi,

I was wondering about possibility of adding a CH3&4 phase inversion to vrecord. As far as I can tell, the "Invert CH2" option only works for CH2, and we recently had a case in which both stereo tracks of a 4-CH recording were inverted. I'd be happy to give this a try myself, but I guess I'm wondering: would it make the most to have different PHASE_VALUE variables for both tracks, to accommodate the rare situation in which one track was out of phase but the other not?

Thanks,

Ben

dericed:
Maybe it needs to be stored as an array of channel numbers, like PHASE_FLIP="1,3" rather than arbitrarily stuck at one value.

privatezero:
Should be resolved by https://github.com/amiaopensource/vrecord/pull/634

bturkus:
@privatezero @dericed thank you so much for doing this! I'm sorry I failed on the testing front (too much art school)!

PR #634, opened 2020-10-29
I'm going inverted

privatezero:
Addresses https://github.com/amiaopensource/vrecord/issues/632 - could use some testing - I'll do some more on my end as well, but have only been able to use virtual inputs so far.

inverted

### DIFF
--- a/Resources/vrecord_functions
+++ b/Resources/vrecord_functions
@@ -79,8 +79,11 @@ _update_config_file(){
 }
 
 _validate_form(){
-    if [ "${INVERT_PHASE}" = "true" ] ; then
+    if [ "${INVERT_PHASE_2}" = "true" ] ; then
         echo "WARNING: Option to invert phase of second audio channel has been selected.\n"
+    fi
+    if [ "${INVERT_PHASE_4}" = "true" ] ; then
+        echo "WARNING: Option to invert phase of fourth audio channel has been selected.\n"
     fi
     if [ "${VIDEO_CODEC_CHOICE}" = "FFV1 version 3" ] && [ "${CONTAINER_CHOICE}" = "MXF" ] ; then
         echo "ERROR: Incompatible video codecs and containers have been selected.\n"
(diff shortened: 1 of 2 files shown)


dericed (approved):
LGTM, but i added to this PR, so back at you

privatezero (commented):
No reason from a code perspective, _report works fine on my end. I think that maybe from a user perspective, _report's output isn't quite as obvious (at least with default colors on my shell)

image (using _report)
 vs.
image (prior to change)

What do you think?

dericed:
oh, thx for showing the difference, maybe in the _report function we could have -w and -W when we need to emphasize

privatezero:
@dericed, added an option to _report that gives red background with white text on my end - should probably test and make sure it gives same outputs on your terminal.

retokromer:
I personally would strongly suggest to **not** increase the possibilities. On the contrary, I guess, it would be better to reduce, in order to make our software better suitable for (partial) colourblind people. In my opinion, the focus on the different possibilities should always have a good **contrast** between background and text. The use of regular font and bold could be a possibility to differentiate different steps of warning or error, if really needed.

privatezero:
Hi @retokromer, the current tweaks to alert color are an attempt to retain the exact same output as is in the current vrecord release, which is much higher contrast than if the `_report` function is left unmodified.

@dericed simplified the code by shifting that particular warning to use the standardized `_report` function (the function that is used for pretty much all the other warnings). That had the unfortunate effect of actually significantly decreasing the contrast for this alert (without expanding the function the warning is just red text), so I'm working to change the function to maintain the current output. Unfortunately, so far the latest tweak results in high contrast on my terminal but poor contrast on macOS, so this solution will have to be improved a bit.

retokromer:
@privatezero Indeed! My remark was meant to go even a step further, suggesting to consider colour impaired people. On my end, it’s not just an academic exercise. There is a person in our staff, who cannot benefit entirely from the colour use in `vrecord` (of course he can consider just the actual text).

dericed:
I suppose an issue could be opened on use of color, but I don't consider this blocking the PR about handling inversion of audio channels.

retokromer:
No, it’s not blocking at all on my end. My remarks came to mind, because we often discuss this issues at our lab.

privatezero:
Merging this - will have to rethink some warning in general, but that seems out of scope of this PR

PR #637, opened 2020-11-12
add dvcont install

privatezero:
add install command that enables dv control functions for linux


dericed (approved):
nice nice nice
not testing, just cheering this on

Issue #639, opened 2020-11-17
vrecord exited a form in an unexpected way (P1_ERR=1) & (P1_ERR=123)


NMAAHC:
```

/Users/medialab
≈:≈ vrecord -e
Summary: FFV1 version 3/matroska yuv422p10 file from ntsc sdi analog. Frame MD5s=Yes, QCTools XML=Yes, concurrent with recording, and Technician=Blake. Inputs recorded to /Users/medialab/Desktop/BC/TR2019-86_5_2_1_2a.mkv and Auxiliary Files created in /Users/medialab/Desktop/BC
Hit enter to start recording

Close the playback window to stop recording.
ffmpeg version 4.3.1 Copyright (c) 2000-2020 the FFmpeg developers
  built with Apple clang version 12.0.0 (clang-1200.0.32.2)
  configuration: --prefix=/usr/local/Cellar/ffmpegdecklink/4.3.1_2 --disable-shared --enable-version3 --cc=clang --host-cflags= --host-ldflags= --enable-gpl --enable-libfreetype --enable-libmp3lame --enable-libopenjpeg --enable-libopus --enable-libsnappy --enable-libtheora --enable-libvorbis --enable-libvpx --enable-libx264 --enable-libx265 --enable-libxvid --enable-libfontconfig --disable-libjack --disable-indev=jack --enable-nonfree --enable-decklink --extra-cflags=-I/usr/local/include --extra-ldflags=-L/usr/local/include
  libavutil      56. 51.100 / 56. 51.100
  libavcodec     58. 91.100 / 58. 91.100
  libavformat    58. 45.100 / 58. 45.100
  libavdevice    58. 10.100 / 58. 10.100
  libavfilter     7. 85.100 /  7. 85.100
  libswscale      5.  7.100 /  5.  7.100
  libswresample   3.  7.100 /  3.  7.100
  libpostproc    55.  7.100 / 55.  7.100
: Device not configured
qcli dev build, Nov  6 2020 02:55:13
filters selected: signalstats aphasemeter astats ssim 

analyzing input file... -
invalid input, analyzing aborted.. 
 2020-11-17T17:52:01 - vrecord exited a form in an unexpected way (P1_ERR=1), please report this error to https://github.com/amiaopensource/vrecord/issues
 2020-11-17T17:52:01 - Error details: /usr/local/opt/ffmpegdecklink/bin/ffmpeg-dl

/Users/medialab
≈:≈ vrecord -e
[...]

PR #640, opened 2020-11-20
exit if no decklink input is found

dericed:
addressing https://github.com/amiaopensource/vrecord/issues/639

Now it just ends with 

```
Hit enter to start recording

Close the playback window to stop recording.
 2020-11-20T09:51:52 - The decklink input is not found.
```


privatezero (approved):
lgtm

PR #644, opened 2020-12-10
Simple GUI

tinytomato:
This code adds a Simple GUI to the Home Page GUI so users can record with saved settings from the home page. It also aligns the buttons to the top, resizes them, and adds a function to vrecord_functions to get the settings summary for the home page.

Questions: 

- Whether to keep the record and passthrough buttons on the bottom; there shouldn't be a functional difference between the two.
- Check out the _get_summary function to see if anything should be added to that report (apologies; my screenshot doesn't show those accurately).
- Should any other setting be added to the simple GUI?
- Not quite related but @privatezero mentioned the Help button no longer works - thoughts?

Screen Shot 2020-12-10 at 3 23 49 PM


privatezero:
Hi @tinytomato, I was able to test this a bit on my end and I think it looks great! The display of current recording mode etc. was working very well on my end! In response to some of your questions:

* I think there should still be different buttons for record and passthrough, that makes it easier for any calibration prior to transfer. Also some people will have different filter views selected for both modes so need different buttons to rapidly access both.
* Not exactly another setting, but I think it would be good to enable the simple gui to load automatically after someone presses cancel in the settings gui. (In the same way that if someone presses 'Passthrough' and then closes the passthrough window that the simple-gui will load again). Seems like that would make usage a little smoother.

privatezero:
Also yeah - we probably should think about what to do with the help button! It is both broken, and the contents of `vrecord -h` which is what it nominally calls are very CLI focused and don't seem to be as helpful for a more GUI oriented workflow. Could that button be replaced with something that would be a little more applicable? A GUI help tab maybe?

tinytomato:
Hi @privatezero - just debugged issues happening in the home dialog not calling the configuration file. Have a look? 
- In response to your question, would deleting the cancel button solve that issue (as clicking "Home" returns the user to the home page). Could also delete the "Home" button and have the cancel button redirect to the home page, but having both redirect seems redundant maybe?

libbyhopfauf:
The only other setting that we change every time as part of the MIPoPS workflow is the recording directory (like going in an creating a new directory to save the video and logs to). Otherwise, this looks fantastic!

tinytomato:
@libbyhopfauf thanks for the comment! I definitely can add recording directory to this GUI. @dericed, @privatezero thoughts?

dericed:
closed and superseded by https://github.com/amiaopensource/vrecord/pull/653

Issue #643, opened 2020-12-10
Remove "setsar=40/27" in recording string


iamdamosuzuki:
Hi all,

This is perhaps a bit of a nitpick, but I think it may be a good idea to remove the `setsar=40/27` and `setsar=16/15` commands (line 1530 and 1544 respectively) from the recording string. 

FFmpeg interprets SAR as "Sample (aka Pixel) Aspect Ratio", rather than what we generally think of as Pixel Aspect Ration. Thus, it appears that this command is simply being ignored, since the PAR (or what FFmpeg calls SAR) is actually being set to 0.9 for NTSC and 1 for PAL. 

Clearly this isn't causing any issues, but I think it could cause some confusion for people trying to interpret the string. Maybe @bturkus could chime in?

dericed:
Are you proposing to leave the SAR unset rather than set?

dericed:
Ah i see, the setsar precedes the setdar, so ya the sar is set twice with setdar overwriting setsar. I agree that removing them would be helpful.

bturkus:
I also agree. This is such a confusing subject in general, and though having setsar in there isn't causing any active harm, since it's being overwritten I could see how this might perpetuate (a) the general confusion about PAR, and (b) the more specific confusion about how FFmpeg operates. And for the PAL people, I believe Morgan meant 1.0667 (16/15); is that correct?

bturkus:
Oh, one more question: are the width and height of vrecord captures are being set by ffmpegdecklink?

iamdamosuzuki:
Sorry about my goof, yeah I meant 1.0667 for PAL (accidentally used 720x570 instead of 720x576 for my calculation, which in a sense if part of this entire problem!)

PR #645, opened 2020-12-11
don't setsar if you're gonna setdar

dericed:
The setsar in these places is pointless since it's overwritten by the subsequent setdar.

to resolve https://github.com/amiaopensource/vrecord/issues/643


retokromer (approved):

privatezero (approved):
lgtm

PR #653, opened 2021-01-23
Consolidate gui

dericed:
Hey @privatezero what do you think of this approach. Here the ui is stacked vboxes with the main dialog (from @tinytomato's work) and the home dialog both being vboxes where the appearance of one hides the other, so they toggle. Still need to add a method for it to return to the last used state, but curious of your opinion of this.


bturkus:
not sure if it's the time for this, but in testing this out I'm getting a variety of errors when trying to capture or enter passthrough mode (trying diff "views" and swapping bet decklink and bmdcapture), whether clicking on the GUI icons or the traditional record/passthrough buttons. 

for ex, "visual" view, decklink selected, vrecord seems to get held up in this way:

```
[decklink @ 0x7fca12803800] Found Decklink mode 720 x 486 with rate 29.97(i)
[decklink @ 0x7fca12803800] Cannot enable video input
UltraStudio HD Mini: Input/output error
pipe:: Invalid data found when processing input

 2021-01-25T10:43:47 - vrecord exited a form in an unexpected way (P1_ERR=1), please report this error to https://github.com/amiaopensource/vrecord/issues
 2021-01-25T10:43:47 - Error details: /usr/local/opt/ffmpegdecklink/bin/ffmpeg-dl
```

dericed:
I'll rebase on main, but the PR is a draft for comments on the arrangement of the gui. Since two guis are merged, I still have work to ensure variable handling works.

dericed:
Hi @privatezero, this is updated to resolve conflicts with main and continue to develop upon the new gui. With the PR, now `vrecord` (no -e or -g needed) would jump to the gui, while `vrecord ID` will try to skip the gui.

privatezero:
@dericed seems to be working well so far on my end, except for a quirk - right now clicking 'passthrough' from the GUI is triggering vrecord to prompt `Enter Identifier:` so some wires look like they are getting crossed

privatezero:
just did a little more testing - the lag time between the passthrough and record buttons being enabled and disabled when switching across tabs is a little bit jarring as well

dericed:
Hey @privatezero, am responding very late to these comments, but I'm having trouble recreating both of them.

With the issue of getting prompted for an identifer in passthrough mode, I'm looking at https://github.com/amiaopensource/vrecord/blob/be704411d80d2e5528c44ce37113f6f130424eb3/vrecord#L2175-L2197 and in either passthrough mode it was play then exit, else if not passthrough mode, it would ask for the id then.

Second one I can't recreate and don't see the lag, maybe an OS specific issue?

privatezero (approved):

PR #654, opened 2021-01-24
Blackmagic dropped Intensity Shuttle

retokromer:
- the product page at Blackmagic does not longer exist
- I suggest to keep the model in our list, because it’s still possible to buy it from various resellers

### DIFF
--- a/Resources/Documentation/hardware.md
+++ b/Resources/Documentation/hardware.md
@@ -6,7 +6,7 @@
   - [UltraStudio](https://www.blackmagicdesign.com/products/ultrastudiothunderbolt)
   - [DeckLink](https://www.blackmagicdesign.com/products/decklink), except for the DeckLink 4K Extreme card (see https://github.com/amiaopensource/vrecord/issues/209#issuecomment-360862657)
   - [Intensity Pro 4K](https://www.blackmagicdesign.com/products/intensitypro4k)
-  - [Intensity](https://www.blackmagicdesign.com/products/intensity)
+  - Intensity
   - [Teranex](https://www.blackmagicdesign.com/products/teranex)
   
   Please note that Blackmagic Mini Converters are not recommended, as they clip the video signal at broadcast levels.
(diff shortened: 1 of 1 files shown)


privatezero:
Hi @retokromer, I'm actually currently looking into if Blackmagic has fully dropped compatibility of USB Intensity Shuttle in Big Sur (which it appears it has). If this is the case, we should probably specify Intensity Shuttle Thunderbolt as being the only option. (The USB version has apparently never worked with Linux).

retokromer:
>The USB version has apparently never worked with Linux

Hmm… I actually used it quite often in Southeast Asia (Ubuntu LTS), as older or simpler computers don’t have a Thunderbolt port.

We have [installed](646) at a client’s `vrecord` on Mac mini (M1, 2020), but not using the Intensity Shuttle, and they are digitising since three weeks. I will check if I can run there a test with Intensity Shuttle Thunderbolt and USB.

privatezero:
Hi @retokromer - I just gave another test with Desktop Video 12.0 (which added Big Sur support for a few more devices) and it unfortunately seems to still be incompatible with the Intensity Shuttle USB on my end

retokromer:
Hello @privatezero, thank you very much for the information! Sorry, I did not had the time nor the possibility to delve into this on my end. However, in my opinion, when Thunderbolt works, then USB must also be possible. I will see if an existing driver can be adapted. (Yet is the Intensity Shuttle still manufactured?)

retokromer:
>(Yet is the Intensity Shuttle still manufactured?)

This PR is about dropping a dead link. It actually opens now the UltraStudio page, but not longer the Intensity one.

privatezero:
Hi @retokromer, the dead link is a problem, but my opinion is that if we are updating this then, for accuracy, we should reflect the latest status of the Intensity Shuttle compatibility.

If there are examples of the Intensity Shuttle USB version being successfully used in Big Sur then just fixing the link is fine, but otherwise I think we should either add clarifying language that Thunderbolt versions are supported, or drop it completely from the list.

(I worry about the possibility of someone a year from now referencing this list and  buying a second hand Intensity Shuttle USB and having it not work for them)

retokromer:
Using the drivers provided by Blackmagic it doesn’t seem to work. And I don’t think there is a solution “out of the box”, I mean without adapting the driver, because Apple did quite many changes in macOS 11 in the way things are handled.

privatezero (approved):
Thanks! It is unfortunate that Blackmagic apparently has so little desire to support their users!

retokromer:
Indeed! I am especially unhappy, because this was a very nice device for simple and efficient video digitisation settings.

PR #655, opened 2021-01-24
include linuxbrew bin path

privatezero:
This fixes a problem with vrecord not correctly finding the path to vrecord resources (at least on my end) when installed via linuxbrew


dericed (approved):
LGTM

retokromer:
If you quote all, it works here without an additional variable.

PR #656, opened 2021-01-24
avoid bmdcapture if using a virtual file input

dericed:
This should allow the virtual input to work even if bmdcapture is selected.


privatezero (approved):
lgtm!

retokromer (approved):

PR #664, opened 2021-02-24
Test ffplay

privatezero:
I think this should help with broken builds of ffplay-dl. Needs some testing though, as my Blackmagic device was bricked! I can take a shot at the DV testing on linux soon, but my apt-install build seems to support libiec61883 so I think it'll work

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -210,15 +210,23 @@ _gather_ffmpeg_vars(){
         BREW_PREFIX="/home/linuxbrew/.linuxbrew/opt/ffmpegdecklink"
     else
         BREW_PREFIX="$(brew --prefix ffmpegdecklink 2>/dev/null)"
+        if [[ -d "${BREW_PREFIX}.reinstall" ]] ; then
+            BREW_PREFIX+=".reinstall"
+        fi
     fi
     if [[ -d "${BREW_PREFIX}/bin" ]] ; then
         FFMPEG_BIN="${BREW_PREFIX}/bin/ffmpeg-dl"
-        FFPLAY_BIN="${BREW_PREFIX}/bin/ffplay-dl"
         FFPROBE_BIN="${BREW_PREFIX}/bin/ffprobe-dl"
     else
         FFMPEG_BIN="$(which ffmpeg)"
-        FFPLAY_BIN="$(which ffplay)"
         FFPROBE_BIN="$(which ffprobe)"
+    fi
+
+    if [[ $("${BREW_PREFIX}/bin/ffplay-dl" ${RESOURCE_PATH}/vrecord_logo.png -t 0.01 -autoexit 2> /dev/null) ]] ; then
+        FFPLAY_BIN="${BREW_PREFIX}/bin/ffplay-dl"
+    else
+        FFPLAY_BIN="$(which ffplay)"
+        FFPLAY_NOT_DL='Y'
     fi
 }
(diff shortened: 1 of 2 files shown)


privatezero:
Just successfully confirmed that this DV play with this change performs as expected (in Ubuntu 20.04 at least)

vrecord
```diff
@@ -210,16 +210,24 @@ _gather_ffmpeg_vars(){
...
         FFMPEG_BIN="${BREW_PREFIX}/bin/ffmpeg-dl"
-        FFPLAY_BIN="${BREW_PREFIX}/bin/ffplay-dl"
         FFPROBE_BIN="${BREW_PREFIX}/bin/ffprobe-dl"
     else
         FFMPEG_BIN="$(which ffmpeg)"
-        FFPLAY_BIN="$(which ffplay)"
         FFPROBE_BIN="$(which ffprobe)"
     fi
+
+    if [[ $("${BREW_PREFIX}/bin/ffplay-dl" ${RESOURCE_PATH}/vrecord_logo.png -t 0.01 -autoexit 2> /dev/null) ]] ; then
```
dericed, line 225: should check `-d "${BREW_PREFIX}/bin"` first (similar to line 217). It's possible for 212 to return a null value for BREW_PREFIX

vrecord
```diff
@@ -462,6 +470,8 @@ _setup_vrecord_process(){
                     VRECORD_STEPS="2" # Steps: record | player
                 elif [[ "${MEDIA_PLAYER_CHOICE}" = "mpv" ]] ; then
                     VRECORD_STEPS="2" # Steps: record | player
+                elif [[ FFPLAY_NOT_DL = 'Y' ]] ; then
```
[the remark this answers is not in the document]
privatezero: thanks! lol I suck at switching between ruby and bash variables

vrecord
privatezero: If brew prefix returned null, wouldn't this still achieve the desired result of throwing a non zero exit status and causing `FFPLAY`_BIN to be assigned to `"$(which ffplay)"`?

dericed: mmm, fair enough

dericed (approved):

retokromer (approved):

retokromer:
@privatezero If it makes troubles on your end, should we modify the ffmpegdecklink formula? I am asking, because we never did experience this issue.

privatezero:
Hi @retokromer, I would support modifying the ffmpegdecklink formula!  I think that a more robust fallback like this is useful for helping account for differences across linux setups, but since we are still using brew for our main source that it would be the best to have the formula create as portable of a build as possible.
I suppose we should at least incorporate the 
>   on_linux do
    depends_on "libxv"
  end

from the main brew formula? (Although this didn't seem to be the root of current problems).

retokromer:
@privatezero The `libxv` “thing” was an issue on our end, but I don’t remember right now if it was related to `vrecord` or not. I’ll try to delve into `ffmpegdecklink` on Sunday and open a PR. (Regarding this PR, I’m very fine with the fallback and I did approve it.)

privatezero:
Great - thanks! I should also note that this issue is affecting both ffmpeg-dl and standard ffmpeg as installed via linuxbrew, so it doesn't seem to be an issue purely with our modified formula

PR #675, opened 2021-07-01
Configuration window entry 'Technician' from config file is not shown…

takahe:
and 'Name of Recording' of config window is not saved to config file


dericed:
closing in favor of https://github.com/amiaopensource/vrecord/pull/676

Issue #603, opened 2020-10-06
horizontal position cross-hairs


bturkus:
Wondering about the possibility of adding some sort of visualization to vrecord that could help with the process of setting the horizontal position of the video. Assuming the positioning is consistently off-center, I try to make adjustments using a TBC, but often, when I eyeball it, using a band of yellow broadcast range pixels as my guide, the resulting sometimes isn't as well positioned as it could be. Is there any chance we could add something like cross-hairs, crop detect, or corners to make this easier?

Thanks!!

dericed:
Would something like the 'corners' view in qctools help with this?

bturkus:
yeah i think it could. Would that be easier to implement/potentially more useful than CropW (in your opinion)?

<img width="1147" alt="Screen Shot 2020-10-06 at 11 15 04 AM" src="https://user-images.githubusercontent.com/8987230/95224711-05ad5a80-07c9-11eb-828c-42be833dd4ac.png">

dericed:
Would it be preferable to have something less instantaneous while calibrating this? For instance, I'm thinking of how in qctools 'lines over time' it shows the same row of video pixels stacked.

PR #677, opened 2021-07-30
add frame positioning video for horizontal calibration

dericed:
resolves https://github.com/amiaopensource/vrecord/issues/603

### DIFF
--- a/Resources/Documentation/settings.md
+++ b/Resources/Documentation/settings.md
@@ -105,6 +105,8 @@ Vrecord will then prompt you for a unique ID. The ID that you type in will becom
 
 **Bit Planes mode** — Bit Planes mode allows to display the video according to the bit position of each plane.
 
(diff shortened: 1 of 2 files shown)


bturkus:
thanks for working on this! seems really great, but a few thoughts:

- wondering if we could change the default background color from black to something like yellow or cyan. tired this out with smpte bars, and the black bar on the far right made it difficult for me to determine the best position. see here:
Screen Shot 2021-08-02 at 3 24 19 PM

- hard for me to describe exactly what's going on, but in testing I found that I was only able to get passthrough mode to work correctly one time. after closing and quitting vrecord, when I would relaunch and select passthrough, ffplay wouldn't pop up. I think something is getting held up somewhere, but can't quite figure out why this is occurring.

dericed:
Hey @bturkus, the ffplay issue is fixed in https://github.com/amiaopensource/vrecord/pull/681 (I hope). I changed the labels to dark gray for clarity.

bturkus (commented):
lgtm sorry for the DELAY

bturkus (approved):
approved!

Issue #686, opened 2021-09-02
vrecord does not detect DeckLink card on Big Sur


iamdamosuzuki:
Hi, I'm unable to get vrecord to detect my DeckLink Extreme 4k on Big Sur. I can get Black Magic Desktop Video to see and use the card properly, but vrecord can't see it. 

<img width="1202" alt="bigsurvrecord" src="https://user-images.githubusercontent.com/3967767/131925179-75bbd78d-ce62-4584-81da-7941c3856bdf.png">

<img width="912" alt="bigsurbm" src="https://user-images.githubusercontent.com/3967767/131925404-7eaec644-59eb-46df-8be3-f3747c6dc4ae.png">

PR #687, opened 2021-09-03
Fixing a homebrew path issue

dericed:
should fix https://github.com/amiaopensource/vrecord/issues/686. Since homebrew installs moved from /usr/local to /opt/homebrew, vrecord is missing a bunch of resources when installed. Need a release after merging this.


privatezero (approved):
can't test, but code lgtm

PR #695, opened 2021-09-22
Update docs

privatezero:
* Added basic Audio Mode page - still needs descriptions of filters
* Tried to scrub as many references to `vrecord -g` as I could since that is now just `vrecord`
* changed short Linux section to crosslink
* updated config screenshot

It's a start anyway


bturkus (approved):
this all looks great thanks for the start!

Issue #700, opened 2022-01-05
Audio + Video with PAL input


privatezero:
Audio + Video mode seems to possibly have issues when using a PAL input rather than NTSC. Needs to be tested with a couple more tapes and fixed if needed!

bturkus:
I can confirm this (and also apologize for not reporting it months ago!)

privatezero:
Thanks for the confirmation! That filter is stretched pretty thin, so not surprised that something went a little wonky - I'll look into seeing what will make it happy.

privatezero:
Resolved in https://github.com/amiaopensource/vrecord/pull/702#event-5867589278

PR #702, opened 2022-01-07
change fr for bars

privatezero:
This should fix the filter chain issues with PAL source and Audio + Video (https://github.com/amiaopensource/vrecord/issues/700) mode while previewing sources.


dericed:
Should we remove the rate attribute altogether instead of hardcoding it?

privatezero:
@dericed the attribute is important to have, as issues seem to arise when that filter runs at 25 fps, and the default for that filter is 25 fps.

dericed (approved):

Issue #705, opened 2022-01-11
Inconsistencies with Preview button


privatezero:
When vrecord GUI is launched via `vrecord` the Preview button will instead launch record mode and prompt for input. Order of conditionals and functions around [line 2240](https://github.com/amiaopensource/vrecord/blob/main/vrecord#L2240) needs some tweaking.

libbyhopfauf:
@privatezero I was just about to post separately about this (but it's an older issue that I've been meaning to create an issue for a bit), but maybe it's related? If not, I can create a new issue:

At MIPoPS all of our computers have a shell script to launch the vrecord GUI (literally just `vrecord -g`) but recently each one had to be changed to vrecord -e to just launch the edit mode instead of the previous set of 6 buttons. If `vrecord -g` is run manually in the terminal,  I get "bad option":

<img width="882" alt="Screen Shot 2022-02-01 at 2 43 56 PM" src="https://user-images.githubusercontent.com/13877631/152063990-0907a608-fab2-4ab1-8643-5f678eefde8c.png">

PR #706, opened 2022-01-13
change setting of record vs edit

privatezero:
This fixes https://github.com/amiaopensource/vrecord/issues/705, but should be tested on other people's side to make sure it causing errors in other contexts!

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -7,7 +7,6 @@ VERSION="2022-01-10"
 SCRIPTNAME="$(basename "${0}")"
 SCRIPTDIR="$(dirname "${0}")"
 CONFIG_FILE="${HOME}/.${SCRIPTNAME}.conf"
-RUNTYPE="record"
 SAT_OUTLIER_THRSHLD=14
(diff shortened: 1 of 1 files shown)


vrecord
```diff
@@ -2237,11 +2243,6 @@ elif [[ "${RUNTYPE}" = "audiopassthrough" ]] ; then
     exit 0
 fi
 
-# record mode
-if [[ -z "${ID}" ]] ; then
-    _report -d "Starting the vrecord GUI. To bypass the GUI, run like 'vrecord FILENAME'."
-    _edit_mode
-fi
 while [[ -z "${ID}" ]] ; do
```
privatezero, line 2246: I think this code might also be able to be removed as with the change in behavior to load GUI by default there seems to be no way to trigger it?

privatezero:
Just tested this on a mac station with passthrough and a short capture and appeared to work in this context as well

privatezero:
This good to be merged?

dericed (approved):

retokromer (approved):

PR #711, opened 2022-02-03
Disk warnings

dericed:


privatezero:
A couple of issue with this on linux - the warning is triggered even with a local volume. Also, it seems to make the default GUI window change in size, which makes the 'Errors and Warnings' bar not display correctly until the window is manually resized.

privatezero (approved):

retokromer (approved):
Tested on Debian 11.1.

PR #713, opened 2022-02-03
Expand language support

### LINKED ISSUE #709: EXIT=audiopassthrough
Hello,
I have got 2 issues.
First : in the vrecord audio tab, the "audio device" column is empty while there are two audio cards.
Second : in the home menu, when i load "Audio Check" Vrecord exits and sends the message 
"numerisation@numerisation:~$ vrecord -e
 2022-02-02T10:18:10 - vrecord exited a form in an unexpected way (EXIT=audiopassthrough), please report this error to https://github.com/amiaopensource/vrecord/issues
numerisation@numerisation:~$ "
My system is Ubuntu.20.04 in a Precision Dell Tower with an Bmd Intensity Pro 4K B inside.
Thanks for help,
JB B

privatezero:
This accounts for the output of audio device parsing via `arecord -l` in linux being multi-lingual.  Will still need some work as non-roman characters (i.e. カード) still don't seem to be parsing correctly, but should fix https://github.com/amiaopensource/vrecord/issues/709 and hopefully accounts for a lot of potential users. Mac functionality doesn't appear to be impacted by language.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -208,13 +208,23 @@ _setup_env_variables(){
 _get_lang(){
     #detects system language
     languages=( de en es fr hi id it jp nl sv vi zh )
-    language_scan=$(locale | head -n2)
+    language_scan1=$(locale | sed -n -e '1{p;q}')
     for i in "${languages[@]}" ; do
-        if [[ "${language_scan}" == *"${i}"* ]] ; then
+        if [[ "${language_scan1}" == *"${i}"* ]] ; then
             system_language="${i}"
             break
         fi
     done
+
+    if [[ -z "${system_language}" ]] ; then
+        language_scan2=$(locale | sed -n -e '2{p;q}' | cut -d ':' -f1)
+        for i in "${languages[@]}" ; do
+        if [[ "${language_scan2}" == *"${i}"* ]] ; then
+            system_language="${i}"
+            break
+        fi
+        done
+    fi
     if [[ -z "${system_language}" ]] ; then system_language='en'; fi
 }
 
--- a/vrecord
+++ b/vrecord
@@ -232,7 +232,7 @@ _set_lang_variables(){
     if [[ "${system_language}" == 'en' ]] ; then
         card_var='card'
     elif [[ "${system_language}" == 'de' ]] ; then
-        card_var='karte'
+        card_var='Karte'
     elif [[ "${system_language}" == 'es' ]] ; then
         card_var='tarjeta'
     elif [[ "${system_language}" == 'fr' ]] ; then
@@ -254,6 +254,7 @@ _set_lang_variables(){
     elif [[ "${system_language}" == 'zh' ]] ; then
         card_var='card'
     fi
+    _include_var card_var
 }
 
 _gather_ffmpeg_vars(){
--- a/vrecord
+++ b/vrecord
@@ -205,6 +205,47 @@ _setup_env_variables(){
     fi
 }
 
+_get_lang(){
+    #detects system language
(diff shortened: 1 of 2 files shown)


vrecord
```diff
@@ -205,6 +205,58 @@ _setup_env_variables(){
...
+        card_var='カード'
+    elif [[ "${system_language}" == 'nl' ]] ; then
+        card_var='kaart'
+    elif [[ "${system_language}" == 'sv' ]] ; then
+        card_var='kort'
+    elif [[ "${system_language}" == 'vi' ]] ; then
+        card_var='card'
+    elif [[ "${system_language}" == 'zh' ]] ; then
+        card_var='card'
+    fi
```
retokromer, line 256: Use a `case` statement rather than multiple `if` statements?

privatezero: I guess that might be a little bit faster, right? (Although I always find `if` easier to read/maintain than `case` for some reason - just a personal quirk I guess).

I think I'd prefer to merge this as is for now, and then maybe convert it to `case` when I circle back to fix the non-roman characters issue and and a few more languages. (Need to do a little more digging to see how many languages alsa tools supports via different words for 'card').

retokromer: LOL, I was actually thinking here about maintenance not speed.

I agree, it’a a matter of taste! I find personally this simpler (and therefore easier to maintain):

```bash
_set_lang_variables(){
    case "${system_language}" in
        en) card_var='card' ;;
        de) card_var='Karte' ;;
        es) card_var='tarjeta' ;;
        fr) card_var='carte' ;;
        hi) card_var='card' ;;
        id) card_var='card' ;;
        it) card_var='scheda' ;;
        jp) card_var='カード' ;;
        nl) card_var='kaart' ;;
        sv) card_var='kort' ;;
        vi) card_var='card' ;;
        zh) card_var='card' ;;
    esac
    _include_var card_var
}
```

vrecord
```diff
@@ -254,6 +254,7 @@ _set_lang_variables(){
     elif [[ "${system_language}" == 'zh' ]] ; then
         card_var='card'
     fi
+    _include_var card_var
```
retokromer, line 257: I guess this should read:

```bash
    _include_var "${card_var}"
```

vrecord
privatezero: lol visually it is much simpler! I think I just hate semi-colons, both in code and in writing so my brain rebels against `case` ;-). That is much cleaner though so I'll probably take your suggestion when I revisit this. I think there are actually still quite a few localizations of alsa-tools that ideally would be  be accounted for.

vrecord
privatezero: I didn't write that [particular function](https://github.com/amiaopensource/vrecord/blob/main/vrecord#L124) but after testing it out, found that indeed using `var` instead of `${var}` is how it was designed:
`var='test' ; _include_var var` will pass `var="test"`to the config file, while 

`var='test' ; _include_var "$var"` will pass `test=""`

retokromer: Interesting indeed, thank you!

vrecord
```diff
@@ -2179,11 +2231,13 @@ fi
 _get_decklink_inputs
 _get_avfoundation_inputs
 _get_avfctl_inputs
+_get_lang
+_set_lang_variables
```
dericed, line 2235: if these two functions are only relevant on Linux for the audio device then can you move then into the linux-conditional in _get_audio_device_list so that they don't unnecessarily run on other platforms.

privatezero:
Closing in favor of https://github.com/amiaopensource/vrecord/pull/714#

PR #714, opened 2022-02-08
grep for brackets in arecord output

privatezero:
Think this might be an easier solution to https://github.com/amiaopensource/vrecord/issues/709. Works on two computers on my end and with output of `arecord -l` examples I can see posted online.

Doesn't help if we have other grep language issues somewhere, but cross that bridge when we get there?


dericed (approved):
much nicer

privatezero:
Nicer if it is in fact universal anyway!

PR #739, opened 2023-03-02
Linuxbrew has been integrated into Homebrew

retokromer:
Question: Is the mentioned RPM install still maintained by Jonáš Svatoš?

### DIFF
--- a/Resources/Documentation/linux_installation.md
+++ b/Resources/Documentation/linux_installation.md
@@ -3,8 +3,8 @@
 ## About
 There are many possible ways to install the various dependencies of Vrecord on Linux. The main body of this documentation is oriented towards installing on Ubuntu LTS. For Some information about installing on additional distributions see the section at the bottom of this document. 
 
-## via Linuxbrew (tested Ubuntu 21.04 and 20.04.2 LTS)
-The following instructions aim to minimize use of linuxbrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord.
+## via Homebrew
+The following instructions aim to minimize use of Homebrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord.
 
 ### Programs to be installed manually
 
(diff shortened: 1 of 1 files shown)


Resources/Documentation/linux_installation.md
```diff
@@ -50,7 +50,6 @@ The following instructions aim to minimize use of linuxbrew installs for package
  test -r ~/.bash_profile && echo "eval \$($(brew --prefix)/bin/brew shellenv)" >> ~/.bashrc
  ~~~
 
-  - Alternately some issues (such as the sdl2 conflict) have been avoided by adding Linuxbrew lower down in the path order than the Linuxbrew instructions call for (Such as by editing `/etc/environment` to include `/home/linuxbrew/.linuxbrew/bin` after the other $PATH directories.) Your mileage may vary!
```
privatezero, line 53: How come this was deleted outright rather than changed to say `Homebrew` like the other sections?

privatezero:
I haven't ever tested that RPM so I can't say, unfortunately

Resources/Documentation/linux_installation.md
retokromer: Was the SDL2 issue not resolved? I cannot reproduce the error on any of the various different Linux configurations we have here (Debian 11.6, 10.13 and 9.13; Ubuntu 22.10, 22.04.1 LTS, 20.04.5 LTS and 18.04.6 LTS). My intent is to remove obsolete information, but of course if it is not obsolete it should remain.

retokromer:
I will send him an email.

Resources/Documentation/linux_installation.md
privatezero: I'm actually not sure if it is no longer a problem or not! I don't seem to have problems with it on my systems, but I'm unsure how it would behave at this point in a totally fresh/clean install environment

privatezero: We do still have the section farther down about manually removing the brew version of sdl still in the [install directions](https://github.com/amiaopensource/vrecord/blob/1d54fa5f216385737d30d971d9380338d6988d84/Resources/Documentation/linux_installation.md#fix-conflicting-sdl2-dependencies)

retokromer: I reverted this part.

privatezero (approved):
Loogs good - thanks!

PR #758, opened 2023-08-25
Fix timecode scan

dericed:


vrecord
```diff
@@ -350,7 +350,7 @@ _setup_vrecord_process(){
         fi
     }
     # set up input and playback
-    unset GRAB_INPUT
+    GRAB_INPUT=()
     GRAB_INPUT+=("${EXTRAINPUTOPTIONS[@]}")
```
retokromer, line 354: Would `GRAB_INPUT=("${EXTRAINPUTOPTIONS[@]}")` nor be the same, but shorter for both lines?

dericed: That would be the same and shorter.

privatezero (approved):
This all seemed to work on my end, although I didn't have a test tape close at hand that has time code

retokromer (approved):
Works on my end (tested on Debian 12.1, Ubuntu 23.04 and 22.04.3 LTS and on macOS 13.4.1).

PR #797, opened 2024-04-26
add caption debugging viewer

dericed:
presumes captions are on line 1 and 2

https://github.com/amiaopensource/vrecord/assets/762604/dfc5963d-9d4e-4ce5-84ea-e9bad460fee1


dericed:
I added eia608 processing during capture, so an scc sidecar is ready just after ntsc capture. ping @bturkus

iamdamosuzuki:
This is working for me. The viewer can see caption data on line 1 and line 2, and an SCC file is created.

privatezero (approved):
I'm just testing this with a virtual input, but seems to work on my end as well!

PR #798, opened 2024-05-09
Few updates

dericed:


vrecord
```diff
@@ -118,7 +118,7 @@ _maketemp(){
     else
         EXT=""
     fi
-    echo "$(mktemp -q "/tmp/$(basename "${0}").XXXXXX")${EXT}"
+    echo "$(mktemp -q -t "$(basename "${0}")")${EXT}"
```
privatezero, line 121: this change is breaking on my end (it throws a 'too few X's' error). Does this work for you?

privatezero (commented):
The change in `_maketemp` breaks the GUI on my end - not sure if this is a linux thing or something else

vrecord
privatezero: Been using that in mac when I've had the occasion - just gave it a shot in Ubuntu and it got a little weird. It's created some packaged files, but it does throw that error and it also creates an empty file when run that is called `"dvpackager: Can't create temp file, exiting....ffmetadata"`. It also appears to be trying to write that same message to chapter markers - I've attached a verbose output here. 
[dvpackager_verbose.txt](https://github.com/amiaopensource/vrecord/files/15300889/dvpackager.txt)

dericed:
I just dropped the maketemp commit. Will get back to that later.

privatezero (approved):
lgtm

retokromer (approved):
Works on Debian 12.5.

Issue #689, opened 2021-09-20
Are we using the correct PAR settings?


iamdamosuzuki:
Hi, I recently had a convo with somebody on the Black Magic forums about the DAR/PAR/SAR of our files. The dude was a bit curt, but his concern seems to be that we're not using the correct values for these. You can see the post [here](https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=147365).

Just to be clear, I'm using PAR to refer to PIxel Aspect Ratio, where a value of 1 would be square pixels. I know that FFmpeg uses PAR to refer to what we generally called SAR (size aspect ratio), which is the ratio of the pixel dimensions of the frame. arggh

The argument is that in D1 video files should have the following attributes:

Frame Size: 720x486 / SAR: 720/486
Clean aperture (active video): 704x486
DAR: 4:3
PAR: 10/11

This is what vrecord uses:
Frame Size: 720x486 
Clean aperture (active video): 720x486
DAR: 4:3
PAR: 9/10

My understanding is that what we're doing is correct because we got a 4:3 DAR, and we have the entire frame as active video since we don't want to crop and potential content and can't control the size of the frame that's coming in from the tapes. Is this correct? Should we have an option to capture the other way?

privatezero:
I'm a little confused by his assertion `Even if you use 486 then you should not change PARs as those 6 lines should be "forgotten"`, as if we used the PAR for 720x480 and were generating 720x486 files then we would be at a 400:297 DAR rather than a 4:3 (if I am doing my sample commands right).

So yeah - is his argument that we should be cropping to 720x480 and then using 10/11 for PAR?

iamdamosuzuki:
Yeah his argument is that we should be cropping to 486, and I think most archivists would argue that we don't want to do this. However, it might be worth asking the community whether or not they want this as an option. 

More interesting is that he claim the PAR of 9/10 is made up and doesn't exist. I guess the claim is that there has never been any sort of standard or best practice that used a 9/10 par to make a 720x486 frame size 4:3. I find this a little hard to believe, but if it's true then we're imposing a new format into an already mixed up situation. That said, I think the 9/10 makes perfect sense as far as access, but is it causing problems for long-term preservation?

alexhabgood:
Hiya, 

I mainly work with PAL, but this article has always provided the best explanation of PAR issues, to my knowledge: https://web.archive.org/web/20110926221502/http://lipas.uwasa.fi/~f76998/video/conversion

[...]

PR #799, opened 2024-05-13
Aspect ratio

dericed:
This address https://github.com/amiaopensource/vrecord/issues/689. Some issues:

It's a breaking change. vrecord used to set both ntsc and pal to 4/3 (no 16/9 or other options). To force NTSC 720x486 to 4/3 used a PAR of 8/9 and for PAL at 720x576 would use a PAR of 16/15. In this version when the user wants a DAR of 4/3, then the PAR is 10/11 for NTSC and 12/11 for PAL. When the user wants 16/9, then it's a PAR of 40/33 for NTSC and 16/11 for PAL.

The default is set to 4/3 even if the user doesn't pick one.

I have a list of unrecommended PAR values at https://github.com/amiaopensource/vrecord/issues/689#issuecomment-2108256577, including the one that vrecord used before this PR, but I hesitate in putting unrecommended PAR values as user options.

I tried to look for standards on NTSC and PAL PAR, and bt601 seems to potentially be that, but it seems quite ignored by the video community, see https://en.wikipedia.org/wiki/Pixel_aspect_ratio#Pixel_aspect_ratios_of_common_video_formats.


dericed:
Here's a picture of the adjustment to the interface. 
image

privatezero (commented):
Bit of a typo in this and the following error message I think

privatezero:
Haven't been able to live test this (my main video station has been apart for some repair work) but the code looks good to me and I agree with general conversation on https://github.com/amiaopensource/vrecord/issues/689

privatezero (approved):

dericed:
Here's the look of SAR tuning under the config menu 

image

dericed:
Lots of updates in here. Now on the main decklink page, is the 4/3 or 16/9 switch, but under config are the fine-tuning options for the 4 combinations of 4/3&16/9 with NTSC&PAL. I think I got all the SAR options discussed in https://github.com/amiaopensource/vrecord/issues/689.

iamdamosuzuki:
This looks great. I haven't had a chance to test this but the screenshots of the settings windows look great.

PR #801, opened 2024-06-18
finish swapping avfctl for dvrescue

dericed:
do not merge until after next dvrescue release

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -949,7 +949,7 @@ if [[ "${OS_TYPE}" = "linux" ]] ; then
         DV_FF_CMD='dvcont ff '
         STATUS_CMD='dvcont status '
     elif [[ "${OS_TYPE}" = "macOS" ]] ; then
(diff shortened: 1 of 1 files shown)


dericed:
dvrescue release is out, this is ready for review

privatezero (changes requested):
I just tested this out and I think that since we have shifted to dvrescue for our main capture mechanism on Linux as well, we should probably drop the system check/conditionals for `dvcont` here and just use the same deck control commands across systems.

privatezero:
The one wonky thing I've found is repack doesn't work as expected on my end (although underlying command works OK when run outside of vrecord). I wonder if this is a shell version difference or something - does this work on your end?

privatezero (approved):
Hmm actually on second thought, it looks like it is throwing some weird messages into the CLI after exiting the GUI. I say let's just go ahead and merge this and I can maybe circle back later to look at the causes and confirm there are no big issues ditching dvcont.

PR #800, opened 2024-05-17
And an mp4

dericed:
a draft towards resolving https://github.com/amiaopensource/vrecord/issues/181. I'd rather not make a ton of options for the second mp4, but gather some consensus on a set of encoding options. Currently I just added `-movflags write_colr+faststart -crf 21 -pix_fmt yuv420p -c:v libx264 -c:a aac` but could mimic the audio mapping of the main output and maybe add some metadata.

I treated the mp4 output as a sidecar file so the option to enable it is at 
image

PR #802, opened 2024-06-18
increase the initial reserved space in the mkv

dericed:
Increasing to 10MB. In https://github.com/amiaopensource/vrecord/pull/800#issuecomment-2176813095 @iamdamosuzuki found that a 90 minute tape needed 3.5MB just for the Cues, so I think increasing the reservation to 10MB would be a save limit.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1604,7 +1604,7 @@ _lookup_choice(){
         "Matroska")
             EXTENSION="mkv"
             DV_EXTENSION="mkv"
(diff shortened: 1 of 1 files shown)


privatezero (approved):
lgtm

iamdamosuzuki:
Started a test capture, it'll run over night. I did get this error when I started the record:

```
Error: 10/11 is not a valid option.
Error: 40/33 is not a valid option.
Error: 12/11 is not a valid option.
Error: 16/11 is not a valid option.
```

PR #803, opened 2024-06-18
sar settings don't need a review

dericed:
This should fix the warnings at https://github.com/amiaopensource/vrecord/pull/802#issuecomment-2176885924.


iamdamosuzuki (approved):
Works great.

PR #807, opened 2024-06-27
ffmpeg verbosity tweaks

dericed:


iamdamosuzuki (approved):
Looks good. Those libx264 messages are gone now.

PR #814, opened 2024-07-17
Allow alternate installs

privatezero:
* Adds option to clear environmental variables from CLI
* Fixes bug of cleared variables not displaying after reset
* Adds option in CLI and GUI to set an alternate directory source for some dependencies to support custom and non brew installations (Currently this can be used to set ffmpeg tools and gtkdialog, but could be expanded)

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -917,6 +917,19 @@ OPTIONAL_TOOLS_GUI=$(cat << CONFIG_FORM
                 <action type="exit">env_reset</action>
             </button>
         </vbox>
+        <vbox>
+            <text>
+                <label>Alternate Dependency Source</label>
+            </text>
+            <entry>
+                <default>"${DEPENDENCYDIR}"</default>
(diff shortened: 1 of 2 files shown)


vrecord
```diff
@@ -2257,9 +2286,11 @@ while getopts ":hc:erpaxvI:O:i:G" opt ; do
         p) RUNTYPE="passthrough" ;;
         a) RUNTYPE="audiopassthrough" ;;
         x) RUNTYPE="reset" ;;
+        n) RUNTYPE="resetenvironment" ;;
```
dericed, line 2289: worth adding to `-h` output?

privatezero: yah - thanks for catching that! I'll add

ablwr (approved):

retokromer (approved):

dericed (approved):

PR #824, opened 2024-08-21
suffixes off by default

privatezero:
Flips suffixes to need enabling rather than disabling as a setting

### DIFF
--- a/Resources/vrecord_functions
+++ b/Resources/vrecord_functions
@@ -74,7 +74,7 @@ _update_config_file(){
         echo "DURATION=\"${DURATION}\""
         echo "PREFIX=\"${PREFIX}\""
         echo "USER_SUFFIX=\"${USER_SUFFIX}\""
-        echo "NO_SUFFIX=\"${NO_SUFFIX}\""
+        echo "YES_SUFFIX=\"${YES_SUFFIX}\""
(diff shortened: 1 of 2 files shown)


Resources/vrecord_functions
```diff
@@ -145,19 +145,20 @@ _get_output_filename(){
...
-        esac
+        if [ -n "${USER_SUFFIX}" ] ; then
+            SUFFIX="${USER_SUFFIX}"
+        elif [ "${DEVICE_INPUT_CHOICE}" = "0" ] ; then
+            case "${VIDEO_CODEC_CHOICE}" in
+                "FFV1 version 3")   SUFFIX="_ffv1" ;;
+                "JPEG2000")         SUFFIX="_j2k" ;;
+                "ProRes")           SUFFIX="_prores" ;;
+                "ProRes (HQ)")      SUFFIX="_prores" ;;
+                "h264")             SUFFIX="" ;;
```
retokromer, line 158: I would suggest `SUFFIX="_ h264"`

retokromer (commented):
One suggestion, otherwise when reading it the code seems correct to me (untested).

iamdamosuzuki:
This works great!

retokromer (approved):

dericed (approved):

PR #825, opened 2024-08-23
Caption updates

dericed:
- avoids keeping an empty file if there's no captions to process
- adds options to make an srt and vtt from the scc


privatezero (approved):
untested with capture, but GUI renders OK on my end

PR #832, opened 2024-08-26
change xpath for slice count

privatezero:
Changes mediaconch policy to use different path for slice count


iamdamosuzuki:
This worked for me

dericed (approved):

PR #841, opened 2024-10-18
Add vplay

bturkus:
Hi all,

I’ve created this branch to introduce vplay.sh, a new script designed to offer playback of already digitized files with the mpv-based QC visualizations that we use during active digitization in vrecord. I wanted to include a few notes regarding this addition:

1. Out of Scope? If anyone feels this is out of scope for vrecord, I’d be happy to remove it. I understand if this doesn’t align with the project’s current direction.

2. Value of the Tool: I do think a tool like this could be valuable, given that vrecord already has the infrastructure for a variety of QC visualizations. While QCTools is excellent, there are situations where the responsiveness of MPV can be particularly useful for quick checks and visualizations.

3. Leveraging the QCView Lua Script: I wrote this with the intention of leveraging the existing qcview lua script. However, we could consider expanding this further, including using the passthrough configuration setting that a user has selected. What I’ve committed so far was significantly easier to implement but could serve as a base for further enhancements.

4. Naming Consideration: I know that the term 'vplay' has been discussed previously in relation to potential tools for testing and evaluating signal routes. If the name or this direction doesn’t align with those plans, I’d be more than happy to set this aside or adjust the approach as needed.

Looking forward to your feedback!


retokromer:
Quick first answer: in my personal opinion, _vplay_ would be a good name, in addition to the _vtest_ we already have.

libbyhopfauf:
@bturkus haven't had a chance to look at this very closely yet, but would this include the ability to go frame by frame? I think this sounds great!

privatezero:
@bturkus - I also haven't had a chance to review closely, but at the surface I think this name and functionality are in-scope enough!

My one thought is that when this gets merged we will have to remember to update the homebrew formula to install it to path when we do the next release

vplay
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
retokromer, line 1: I suggest to use the generic shebang here as well:
`#!/usr/bin/env bash`

dericed (approved):
APPROVED

PR #846, opened 2024-10-25
mapping for mono playback

privatezero:
Making this PR for testing. This should make all the mono options 'what you hear is what you get' in terms of channel mapping in monitoring and file output.

Let me know if people have a chance to test/have any thoughts about this! This will make it harder for people to accidentally have vrecord set to capture the wrong channel for mono, but also potentially would allow people to miss realizing a second track exists when vrecord is set to capture a single mono track, as the second track will no longer appear in the meters.

The only way I can see to avoid this is to have two sets of audio meters, one that displays volume for all tracks and one that displays volume for mapped tracks...but that probably is then making the views overcrowded.


dericed (approved):

bturkus (approved):
this looks great to me!

privatezero:
Rad! Thanks you two for checking

PR #865, opened 2025-02-25
Profiles

dericed:


dericed:
@iamdamosuzuki can you suggest what settings should be global vs per-profile

- noting that we need a way to rename a profile

dericed:
perhaps let's make it so you pick the input, then conditionally pick a related profile

iamdamosuzuki:
Possible bug: The profile wont save if an Input Device isn't selected. I noticed this while trying to test this out without an input device hooked up to the computer.

iamdamosuzuki:
Change request: Add the MP4 Options (CRF value and extra options) to the Settings Summary

iamdamosuzuki:
Change request: Add the selected SAR value to the Settings Summary based on the Broadcast Standard and DAR values selected in the profile.

iamdamosuzuki:
Change request: As discussed in our meeting, remove the Input Device selection from the advanced options and have it selectable in the simple GUI

vrecord
```diff
@@ -6,11 +6,37 @@
 VERSION="2024-10-18"
 SCRIPTNAME="$(basename "${0}")"
 SCRIPTDIR="$(dirname "${0}")"
-CONFIG_FILE="${HOME}/.${SCRIPTNAME}.conf"
+PROFILE_DIR="${HOME}/.vrecord_profiles"
```
retokromer, line 9: I would suggest to use
`PROFILE_DIR="${HOME}/.config/vrecord_profiles"`
i.e. inside the “canonical” configuration folder, and “visible.”

retokromer:
@dericed Has this been abandoned?

PR #871, opened 2025-03-04
add Hi59

dericed:
getting this started again. #wip

### DIFF
--- a/Resources/vrecord_functions
+++ b/Resources/vrecord_functions
@@ -125,7 +125,7 @@ _validate_form(){
         echo "WARNING: Option to invert phase of fourth audio channel has been selected."
     fi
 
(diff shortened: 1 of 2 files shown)


bturkus:
Hi @dericed I did some testing this morning, and it seems to be working pretty well. Few questions:

(1) Was there a visualization view that you thought would be best for HD? I mostly tested w/ unfiltered, as some of the others (Audio + Video) didn't seem well-suited to HD
(2) I think we'll need to tweak some MediaConch policy checks for HD
(3) This is the impt one--I found that during playback, the audio was really suffering, getting garbled, disappearing entirely for sections, coming back, etc., though the resulting files all seem fine. I think this is maybe a more extreme version of some of the buffering issues we see with longer captures of SD material?

privatezero (approved):
Can't test rn, but code lgtm

retokromer (approved):

PR #873, opened 2025-03-05
remove ffprobe dependency

dericed:
This is a big dependency for a quick evaluation that could be done with ffmpeg, removing ffprobe for now.


privatezero (approved):
Only have time to check with virtual input this morning, but it seems to work on my end (and looks good to me)

PR #874, opened 2025-03-24
refactor the approach to mediaconch policies

dericed:
Now we can conditionally build a mediaconch policy


privatezero (approved):
This seems to work for me so far

PR #883, opened 2025-04-03
Added addional HD formats

iamdamosuzuki:
Added in all remaining HD formats compatible with Decklink SDI 4k card:

- 23ps - 1080p23.98
- 24ps - 1080p24
- Hp25 - 1080p25
- Hp29 - 1080p29.97
- Hp30 - 1080p30
- Hi50 - 1080i50
- Hi59 - 1080i59.94 (this was already in there)
- Hi60 - 1080i60
- hp50 - 720p50
- hp59 - 720p59.94
- hp60 - 720p60

Added in description of the format as well, which helps the user. Put this all in the main window and adjusted main window settings slightly so that the formats names and descriptions are readable. I tested each of these with an HD-SDI signal generator and they all worked. Happy to upload test files if anyone wants to see them.


dericed:
A lot of these have no corresponding tape format. Is there a 720 height digital tape out there?

iamdamosuzuki:
I've seen 720 HDV before. But i understand your point. I'm happy to trim this down. I'll look into the HDV and HDCam size/rate lists. I feel like between those two formats a lot of these would be covered,

iamdamosuzuki:
According to [this site](https://www.freevideoworkshop.com/difference-between-hdv-and-dv/) HDV supports the following:

- 720p25
- 720p30
- 720p50
- 720p60
- 1080p24
- 1080p25
- 1080p30
- 1080i50
- 1080i60

According to the [Sony HDW-M2000 manual](https://www.manualslib.com/manual/162738/Sony-Hdw-M2000.html) the VTR supports the following formats (page 10-4)

- 1080i50
- 1080i59.94
- 1080p23.98
- 1080p24
- 1080p25
- 1080p29.97

iamdamosuzuki:
I think it's safe to say it's possible that between HDV and HDCAM each of these formats could exist on a digital tape. That said @dericed, I understand definitely understand the concern here, as these options clutter the interface. What about a checkbox in the config menu that makes HD formats visible in the format selection list? I'm open to whatever people think is best, I mostly wanted to see if this would be possible (and it is!)

privatezero:
Kind of tangentially related, but is analog capture of HDV still our best option? I haven't been able to test since I don't have any HDV around, but on the Linux side of things FFmpeg's [IEC61883 device](https://ffmpeg.org/ffmpeg-devices.html#iec61883) (which is already integrated into vrecord) in theory supports HDV capture. Is this a capability we should explore more?

harrypm:
> Kind of tangentially related, but is analog/SDI capture of HDV still our best option? I haven't been able to test since I don't have any HDV around, but on the Linux side of things FFmpeg's [IEC61883 device](https://ffmpeg.org/ffmpeg-devices.html#iec61883) (which is already integrated into vrecord) in theory supports HDV capture. Is this a capability we should explore more?

DVgrab on Linux (also HDVsplit Premiere and Final Cut) all support MPEG-2 transfer directly to file for all HDV flavors, although a bigger question is handling the firecontrol metdata as atleast the last gen Sony HDV camcorders also recorded all settings changes in real-time if you have ever played back a tape in-camara you will notice that little feature.

iamdamosuzuki:
Hi @harrypm, which versions of FinalCut support HDVsplit? Are there any option for Mac? 

I think that this is a worthwhile conversation, but considering the difficulty of working with this format I think giving people a solid option for HD SDI capture directly to FFV1 is a net win.

harrypm:
Not sure honestly I have a older copy on a MBP 2017, last I checked though they haven't removed the interface feature from the current versions.

But all my HDV stuff is handled with DVgrab on Linux as it's all CLI exactly the same workflow as DV.

It's just MPEG-2, so SDI and FFV1 what makes no practical sense because it will burn a drastically higher amount of storage space for an already digital file stream, all I do is just remux to MKV and PCM to FLAC, persevering the original 25-30mbps 4:2:0 8-bit video stream.

iamdamosuzuki:
Hi @harrypm, i agree that capturing the MPEG-2 stream is the best practice for preservation, I think the reality is that a lot of labs don't have access to the hardware and software needed to in order to do this kind of capture. My thought is that vrecord as a tool should help people preserve tapes as best they can, but not necessarily make prescriptive choices about how to preserve their tapes. So, giving people the option to capture HD SDI streams should be available for folks who don't have other options, with the understanding that for some formats it will inflate the size of the video.

iamdamosuzuki:
btw, I'm working on a checkbox to turn the HD formats on and off in Standard window. I think that strikes a nice balance between overcrowding the interface and giving users all the available options.

bturkus:
@iamdamosuzuki we encountered a 30p HDV tape and tried this out, but when I click the "Show HD Formats" button, the vrecord window disappears/reappears but I'm not seeing any formats pop up anywhere. Is there something I'm missing?

bravoalphatango:
A few things...

Updated to 2025-04-21 but am not seeing the 'show HD formats' button. Am I missing something? Or is this something ya'll are testing out and isn't in this release?

I would like to advocate for a m2t option in vrecord, but I know I am not the person making these changes. My question is: Would it be significantly more work to incorporate this data stream than adding a collection of HD format? @dericed? If the capacity for folks to capture HDV as m2t is incorporated doesn't that eliminate the need to use obsolete hardware and software? If it is simply not something that can be accommodated as free work I get that, but just wanted to clarify if that was the actual issue.

I also wonder who is using vrecord outside of the archival community (who use it in a preservation context and would prefer an m2t file). I could be wrong, but my assumption was that vrecord isn't used much outside the archival community? Please come for me if I am incorrect!

bturkus:
@bravoalphatango 

Thanks for these thoughts—lots of good points here.

Re: the “Show HD formats” option, that’s actually only available in a branch Morgan set up for testing. You can try it out by going to [this branch](https://github.com/amiaopensource/vrecord/tree/add-hd-config), downloading the ZIP, making vrecord executable, and running it with ./vrecord -e from within the folder. It's not in the main release yet.

On the m2t front, I think many of us would love to see an m2t capture option in vrecord. The challenge, as I understand it, is that it’s not a simple lift—it would require significant work to incorporate support for MPEG-2 Transport Streams from HDV over FireWire. That said, @privatezero  recently pointed out some potentially promising FFmpeg developments that might make this more feasible. It’s something that could definitely be investigated further, and I'd be happy to sponsor or co-sponsor, if funding allows.

The current HD approach isn’t ideal—it was more about creating a way to keep HD capture within the vrecord ecosystem. For us, Media Express is the fallback alternative, as we can't deal with keeping an old Mac around just to run legacy m2t software.

privatezero:
@bturkus I don't know if the things are _promising_ for mac users, but vrecord already has the option of capturing via the IEC61883 device, so in theory already could support HDV on Linux - my understanding is that both that device as well as the [dvgrab](https://linux.die.net/man/1/dvgrab) tool that would be an alternate way of facilitating HDV capture both rely on libraries that aren't/won't be ported over to Macs. So yeah...HDV capture in vrecord on Linux is theoretically pretty simple (although I haven't been able to test) but Mac users might be high and dry without some more significant investment in development of the underlying tools

bravoalphatango:
Offered to make @privatezero some test tapes if that would help assist.

privatezero:
@bravoalphatango that would be rad! Yeah, I feel like for the linux side anyway we could get this refined. (Although I know a vast majority of our users are Mac users).

 I have a long standing idea of pitching a 'good vibes guide to setting up a linux transfer station' talk...maybe I need to think about that some more
image

iamdamosuzuki:
Thanks everyone for taking the time to think over this! One think I want to clarify is that I of course would prefer to have m2t support in vrecord, but it's clear that's not something that can be dropped in overnight. The reason I think it's worthwhile to add HD support is because 1) it makes it possible to capture HD directly to FFV1, which isn't possible using MediaExpress and 2) it's extremely easy to do. All the HD formats are natively supported in the Black Magic SDK. So, it gets us part of the way there with very little effort. In any case, I can't get the checkbox to work, but Dave said he'd work on it. It works if you close the window and reopen it. If any one has any ideas on getting it to reload the format list upon clicking the box please be my guest and update it. 

Also, I'd love to have a "good vibes linux setup" document. I've always gone with macs because it just works. But I think getting away from vendor lock-in and having robust linux support would be an amazing next step for this platform. On the other hand, mac minis are cheap as hell to purchase, so at least someone can build a solid capture station for around $1,200 all in ($500 for mac, $350for black magic SDI card, $350 for the pcie enclosure).

bravoalphatango:
This all makes sense. I figured there was some part of the conversation I was missing since this has been a hot topic for so long.

I’d be very into a “good vibes guide to setting up a linux transfer station“. I too work exclusively on Macs but if setting up a Linux station offline (IT won’t support Linux for some reason) enables HDV transfer I’m all for it! I mean we already have a separate station setup for HDV transfer offline anyway with a 2006 Mac Mini.

privatezero:
@iamdamosuzuki just to clarify - I also think it is worthwhile to add the HD support, since it is supported by the hardware! If the checkbox refresh issue gets to be too annoying to fix, maybe just having a "HD Video" option in the main drop down menu, with a different menu on the configuration tab to set which specific flavor of HD capture is used when someone selects "HD Video" might be a work around?

PR #889, opened 2025-04-13
if DIR is empty set to pwd, only test DIR if exists

dericed:


privatezero (approved):
This seems like it works on my end!

PR #876, opened 2025-03-27
Test signal output

privatezero:
Add the option to use a test signal as input via the `-s` option in the command line - addresses https://github.com/amiaopensource/vrecord/issues/867

PR #890, opened 2025-04-14
Aphasemeter to axcorrelate

privatezero:
Switching the phase graph in the audio preview windows to use `axcorrelate` filter rather than `aphasemeter` makes vrecord audio interface behave more in line with behavior of audio meters in other software. (Makes the phase graph independent of channel amplitude).

@dericed - do you have any thoughts about the filter settings for this? It looks like the `best` mode is not yet in the version of ffmpeg we have `ffmpeg-dl` locked to for now.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -2021,25 +2021,25 @@ else
     vectorscope=i=0.04:mode=color2:c=1:envelope=instant:graticule=green:flags=name,\
     scale=400:400"
     if [[ "${AUDIO_MAPPING_CHOICE}" = '2 Stereo Tracks (Channels 1 & 2 -> 1st Track Stereo, Channels 3 & 4 -> 2nd Track Stereo)' ]] ; then
-        PRINT_PHASE=",drawtext=fontfile=${DEFAULTFONT}:box=1:text=Phase\\\: %{metadata\\\:lavfi.aphasemeter.phase}:x=135:y=180:fontcolor=black"
+        PRINT_PHASE=",drawtext=fontfile=${DEFAULTFONT}:box=1:text=Phase\\\: %{metadata\\\:lavfi.astats.1.DC_offset}:x=135:y=180:fontcolor=black"
         PRINT_CHANNELS_1_2=",drawtext=fontfile=${DEFAULTFONT}:text=Ch.1/2:x=10: y=10:fontcolor=white"
         PRINT_CHANNELS_3_4=",drawtext=fontfile=${DEFAULTFONT}:text=Ch.3/4:x=10: y=10:fontcolor=white"
         AUDIO_SPLIT='7[a][aa][b][c][d][e][out1]'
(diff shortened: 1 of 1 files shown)


bturkus:
Hi @privatezero I'm having the weirdest problem, with this AND with the current release of vrecord. When selecting 2 audio tracks, I'm getting strange readings in Audio + Video mode:

Screenshot 2025-04-15 at 9 04 12 AM

Even though I'm sending in 2 CH test tones at -18dB on both tracks, you can see 3&4 are coming in low, and the phase of 1&2 is bizarrely off. I can file a separate issue, but any ideas what's happening?

privatezero:
@bturkus yeah, that seems like something we'll have to follow-up on so a separate issue is probably helpful for keeping things organized. When I was working on the [test signal input branch](https://github.com/amiaopensource/vrecord/pull/876), I noticed some weirdness with levels in the meters for channels 3&4 of a virtual quad track input, but I haven't had time to dive into that all yet. I guess it is time for a deeper dive into how the `pan` filter is interacting with all our outputs across modes.
That's just the level issue though...the phase issue...that just seems weird! If it is happening in the same way using the unrelated `axcorrelate` filters and the `aphasemeter` filter in the existing release, I am not sure what that could be - I can't say I have noticed any phase monitoring issues when I've been testing things in quad...I'll run some more tests though. Does the phase only look wonky on your end when capturing to 2 stereo streams?

iamdamosuzuki:
I'm having similar issues as Ben. I'm sending a 4 channel signal to vrecord but channels 3 and 4 in the monitor are not related to what I'm sending it. However, unlike Ben I'm not having any phase issues. 

Screenshot 2025-04-16 at 7 45 10 AM

What differences should we be seeing between `axcorrelate` and `aphasemeter`? I don't see any significant difference, and besides the Channels 3/4 issue this looks good to me. However, let me know if there's something we should be looking for.

privatezero:
@iamdamosuzuki - the level discrepancy issue is fixed in https://github.com/amiaopensource/vrecord/pull/892 (so part of main now, but not yet in this PR).

The main difference you should see is that the `axcorrelate` graphs should be independent of amplitude of the signals. With the `aphasemeter` generated graphs, if you for example pumped a stereo tone into it and turned down the volume of one of the channels, it would register that as moving more out of phase on the graphs. With `axcorrelate` differences in the amplitude of the channels doesn't affect how it reports the phase information.

Note - this is _just_ for the graphs and numerical phase value part in the upper left of the window and doesn't impact how the `avectorscope` portion works.

iamdamosuzuki:
In that case the testing looks good! I'll approve

iamdamosuzuki (approved):
I tested this out. Changing the volume of the audio channels does not change the phase in the axcorrelate window.

privatezero:
@bturkus and @iamdamosuzuki thanks for the testing! Glad to get this resolved (and to have fixed another problem at the same time!)

PR #899, opened 2025-05-29
Redo output strategy

dericed:
this refactors the output to use the tee muxer (which has the recording output and also the output to the player). With the tee muxer, we can specify how any output stops based on the end of another output, so that helps us get off ffmpeg@5 to ffmpeg@7.

The output now looks like (example here for two stereo tracks in the output): ` -f tee [f=matroska:select=v,a\\:1,a\\:2]WHATUP.mkv|[f=nut:onfail=abort:select=v,a\\:0,a\\:1,a\\:2]pipe:1`. These stream mappings refer to output streams rather than input ones. Elsewhere the first output audio is always a copy of the input (8 channels for the decklink) and then the others are the recording output audio streams, so the recording output gets `v,a\\:1,a\\:2` and the output to the player gets `v,a\\:0,a\\:1,a\\:2`. Next to do is to update the player to show the input vs output audio.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -568,7 +568,7 @@ _setup_vrecord_process(){
         RECORD_COMMAND+=("${MIDDLEOPTIONS_PRES[@]}" "${MIDDLEOPTIONS_ALL[@]}")
         if [[ "${RUNTYPE}" = "record" ]] ; then
             RECORD_COMMAND+=(-filter_complex "[0:v:0]${RECORDINGFILTER#,*}${TC_WRITE}${CAPTION_WRITE}[vout];${AUDIOMAP}" -map "[vout]" "${AUDIO_CHANNEL_MAP[@]}")
-            RECORD_COMMAND+=(-f tee [f=${FORMAT}:select=v,${AUDIO_TEE_SELECT_MAP}]${VRECORD_OUTPUT}\|[f=nut:onfail=abort:select=v,a\\\\:0,${AUDIO_TEE_SELECT_MAP}]pipe:1)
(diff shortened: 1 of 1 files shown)


dericed:
Thanks @iamdamosuzuki, this is really good to hear. In my testing I was only using some file inputs that already had the video tracks in stream order before the audio tracks, but decklink puts the audio first, so it needs to be mapped in a specific order. I'll add that soon. Really glad that mediaconch watches our work as we make changes.

dericed:
Hey @iamdamosuzuki, I added 2 commits. It should pass mediaconch policy now. Could you retest?

iamdamosuzuki:
Yes, I'm testing capture of FFV1/MKV. However I'm also creating qctools and mp4 access files concurrently. 

I just completed an 8 hour long capture of FFV1/MKV + MP4 (no qctools because it won't work for a file that long). I see no av sync errors and found no drop frames (using framemd5).

privatezero:
I've been testing on my end - so far things seem fine except that with the new branch I've noticed an oddity in how Vrecord parses the QC tools metadata to report crushed audio. Basically vrecord will report that all frames have crushed audio levels, even when they don't. Have been testing this with short captures of tone out of our TBC and will see if I can figure out what is causing the issue.

Other than that I've found no dropouts as a result of capturing with this branch, although I did have some issues before upgrading qcli to the latest daily build from the December release - would be curious if anyone else can duplicate that.

iamdamosuzuki:
my 200 min capture went fine with no dropped frames.

@privatezero I'm also getting those crushed audio errors.

dericed:
This last commit `probably` fixed the qcli test error. Before that commit, qcli was processing the same audio that the player gets which is the full 8 channel decklink audio plus all mapped streams. This last commit puts the 8 channel audio last and restricts qctools to only testing the first track (which would be the first of the user mapped ones).

PR #900, opened 2025-06-10
Use a lossless encoding internally when using test signal and passthrough

dericed:
this pr is based on https://github.com/amiaopensource/vrecord/pull/899


dericed:
rebased on main, ping @privatezero

PR #901, opened 2025-06-10
Add recordingfilter to passthrough mode

dericed:
also based off https://github.com/amiaopensource/vrecord/pull/899


privatezero (approved):
I've tested this out with capture on both mac and linux and this seems to be working well in both

privatezero (changes requested):
Oh - holdup, there seems to be a problem with Audio + Visual mode

dericed:
ready to re-review

privatezero (approved):
Seems to work on my end now

PR #907, opened 2025-06-23
switch to ffmpeg-ma

dericed:
waiting on https://github.com/MediaArea/homebrew-mediaarea/blob/master/ffmpeg-ma.rb to add --with-iec61883 but otherwise ready to test

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ This documentation is up to date as of vrecord version 2020-02-10.
 
 Vrecord is open-source software for capturing a video signal and turning it into a digital file. Its purpose is to make videotape digitization or transfer easier. Vrecord can capture analog and digital signals through a variety of inputs and can create digital video files in a variety of formats and codecs. Vrecord has been designed with needs of audiovisual archivists in mind. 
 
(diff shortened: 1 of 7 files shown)


retokromer (commented):
Good move indeed!
Untested, but looking at the code it seems correct.

privatezero:
> ```
>     if OS.mac?
>       args << "--enable-audiotoolbox"
>       args << "--enable-videotoolbox"
>     end
> ```

Yes - conditionals would be how I would handle things. @dericed, do you have any sense as to how willing the maintainers of that formula are to make or accept changes?

privatezero:
with the changes in https://github.com/MediaArea/homebrew-mediaarea/pull/115 this now builds on my end (Ubuntu 24.10)

privatezero:
@dericed I've been doing some testing to see if this branch will resolve the h.264 encoder issues reported in https://github.com/amiaopensource/vrecord/issues/908, and it looks like switching how we encode h.264 has caused some incompatibilities with our current code. So far I am running into errors for `Unrecognized option 'crf'` and `Unrecognized option 'x264opts'`, so with the shift to using `libopenh264` it would appear we will have to rethink some flags around MP4 creation.

retokromer:
@dericed @privatezero see https://ffmpeg.org/ffmpeg-all.html#libopenh264

privatezero (approved):
Probably will have to do some work in the future to double check that alternate builds/installs outside of brew remain a possibility, but this lgtm! One question though - with the switch over to ffmpeg-ma, do we still need all the brew related conditionals, or is the `$(which ffmpeg-ma)` line enough?

Issue #839, opened 2024-10-17
Display Showvolume Audio Channel Names 


bturkus:
Hi,
I was wondering if people would be open to tweaking the showvolume filter in "Audio + Video" passthrough mode to display the audio channel names. We could do this in a few other places as well, but I think passthrough in A+V mode would be the most beneficial, as novice vrecord users would not necessarily know that CH1 is stacked on top of CH2 (this just occurred in a Video Preservation class--I knew we were wired improperly, but no one else picked up on it).

@privatezero as I believe you implemented this, what do you think?

Thanks,
Ben

privatezero:
Hi @bturkus we can definitely do this as channel names are already part of that filter - I think we turned them off because the levels tend to be low enough on all these old recordings that the channel names kind of interfere with the bars at the far left end of the meter.

The meters are set to linear right now, but we could change them to logarithmic and maybe that would be more useful? It would make the channel names way easier to read at least. (Example below in log mode with channel names on). Anyone have thoughts about going to logarithmic audio meters?
![image](https://github.com/user-attachments/assets/5c7532dd-111f-4651-880d-7a4b02d055ed)

privatezero:
I made a quick branch you can try out to see what you think: https://github.com/amiaopensource/vrecord/tree/logmeters

bturkus:
thanks @privatezero I'm sure this is great, but I'll test on Monday to confirm. Unrelated, but looking over active branches, do you need someone to test changes you made to mono audio configurations (audio-monitoring-tweaks)?

privatezero:
@bturkus oh yeah testing on that branch would be great! Can't remember why I didn't do a PR for that yet, I probably just got distracted with work. I think the main thing I was thinking about was the best way to handle the monitoring when a user selects the option to split two input channels into two separate mono tracks (since in our preview window you can't switch between two different audio tracks on the fly). Right now it just lets the user monitor them in stereo basically, which I think might be the best option but would be happy for any other thoughts about that!

richardpl:
If you do not like transparency of default showvolume, you can add ,format=rgb0 after showvolume filter to force always black.
This will make names more visible, other alternatives is rescaling showvolume filter output with xbr/epx/hqx or similar filter.

privatezero:
@richardpl thanks for those! For myself, I kind of like the transparency on top of the video signal, but I'm curious what others think.

bturkus:
[...]

PR #913, opened 2025-07-09
label the selected audio tracks for output

dericed:
how about something like this for https://github.com/amiaopensource/vrecord/issues/839 ?

For 2 tracks of stereo
image

For 2 channel to 1 mono track
image


dericed:
This is putting M, L, or R for the role of the channel on the output (mono left or right) and then the output audio track number in `()` (counting from 1)

privatezero:
This seems like it should work to help with labeling the active channels - one thought though, doesn't switching to `AUDIO_PLAY_MAP="acopy"` have the effect of reverting us back to all channels being audible in preview as opposed to just channels that will be actively captured?

dericed:
ya the acopy edit was just get the unmapped input audio consistent and then conditionally label that. I guess we need to use that for this visualization but then the old mapped audio for preview.

PR #923, opened 2025-08-14
Fix recording after timecode scan

dericed:
@privatezero here's a few quick fixes when testing on the latest release

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1541,6 +1541,7 @@ _edit_mode(){
         done
         echo "--------------|---------------"
         RUNTYPE="edit"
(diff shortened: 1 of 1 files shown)


privatezero:
@dericed I tested this on my Mac and with the change the mediaconch policy to 32 bit it actually failed, as my audio _was_ coming in as 24 bit rather than 32 bit. I can take a look again tomorrow (as I think I have the same blackmagic model as you?)

privatezero:
Can confirm - this is what I'm getting today (scanned with ffprobe)

`Stream #0:1: Audio: flac, 48000 Hz, stereo, s32 (24 bit)` captured via Ultra Studio HD Mini

Maybe add an `or` line in the flac policy to support either 24 or 32?

dericed:
merged, thx

Issue #810, opened 2024-06-28
Show decoded caption data in Caption viewer


iamdamosuzuki:
I would like to have the option to view the decoded caption data in the Caption Viewer mode, rather than the raw hex values.

dericed:
Hmm, I'm not sure how to get ffmpeg to do this, cc'ing @richardpl for advice.

Currently the caption values as accessible in their hexadecimal form via the metadata filter (as used at https://github.com/amiaopensource/vrecord/blob/e933bd84d62a0bd5edaa8891f6a20557c116443b/vrecord#L551) and then output to a file. Then a function processes that metadata output from those hex values into an scc (https://github.com/amiaopensource/vrecord/blob/e933bd84d62a0bd5edaa8891f6a20557c116443b/vrecord#L265-L340, this is borrowed from sccyou).

To get onscreen captions for a live input, then the output of the metadata filter would need to be sent back to ffmpeg as an additional input. I see that ffmpeg has an 608 decoder but I'm uncertain how to accept that as an input as there's no `-f rawdata -c eia608` options.

richardpl:
You could write .scc files and use .scc demuxer from libavformat, if you instead need realtime/stream solution than its more complex.

iamdamosuzuki:
This has been addressed. At the current time the caption viewer only works in passthru mode.

Issue #499, opened 2020-02-07
Start and stop recording at specified timecodes


YohannLYD:
Being a very lazy technician, I wonder if there would be a way to start and stop a capture the same way Media Express does when we specify the timecode in/timecode out ?

At the moment I see two things that could make it difficult at the moment :
* The low level RS422 commands to control a deck don't seem very well documented by decks manufacturers
* vrecord is single threaded, which would prevent the recording and real-time TC reading being done at the same time

Regarding the deck control the decklink API could be used but vrecord being a bash script, it might be better if we could control it directly sending low level commands to the `/dev/tty.RS422 Deck Control` interface.

For the single thread issue, maybe a side script could read the TC from the serial port and start and stop the recording when the values are reached ?

Just mixed ideas there... :)

Side note if it can be useful, I found the link for an unofficial Sony 9-Pin protocol specs which I was looking for to control our JD30 deck.  
[ftp://ftp.jurassic.nl/mirrors/ftp.sgi.com/sgi/video/rld/vidpage/s9pinnt.html](ftp://ftp.jurassic.nl/mirrors/ftp.sgi.com/sgi/video/rld/vidpage/s9pinnt.html)

dericed:
cc'ing @bturkus on this idea, but we were thinking of merging the functions of https://github.com/bavc/deckcontrol into vrecord. Also the features you're suggesting seem doable. We could have a background process watch the tail of the timecode file as it writes and send a kill command to ffmpeg when a timecode out is found. That would be close but not frame accurate.

bturkus:
Hi @YohannLYD this is great idea, and as Dave kinda alluded, we've been planning on some deck control-related work sometime soon. If you have any additional ideas, please share them here (or in separate issues), and we'll try to incorporate them into future plans. Thanks!

YohannLYD:
I didn't know about deckcontrol. It's a pretty neat utility! Tailing the timecode file sounds like a good solution as perfect frame accuracy is not a requirement in my case.

Is there a basic ffmpeg command line to output the vitc2 timecode value on the stout?  
I am going to give a try to see if I can tweak something between a ffmpeg process and deckcontrol.

PR #935, opened 2025-09-17
[WIP] this adds a live display to the caption viewer

dericed:
this is a draft to get to https://github.com/amiaopensource/vrecord/issues/810. I starts adds the FFREPORT environmental variable and the play or record command adds the metadata filter to print the hex captions there. A background process called cchex_to_display then periodically skims the end of the FFREPORT and uses a simple ASCII lookup table to convert some of the hex caption data into text and then a drawtext filter uses textfile and reload to draw it live. Currently the live captions only include the basic ASCII and not any extended set, there's also no placement or control codes processed. And it is only implemented in the "Captions" viewer thus far.

Since the captions aren't really fully processed, I'm not using the black boxed style as captions go for, but i'm using a shadowed, right-to-left scroll. I tried to do this all in ffmpeg but I'm not sure it's possible without a video->subtitle filter. @richardpl I'm curious of your thoughts.

https://github.com/user-attachments/assets/0de5f8b2-82df-4c0a-a7c8-9d7da7aecb64

### DIFF
--- a/cchex_to_display
+++ b/cchex_to_display
@@ -1,0 +1,170 @@
+#!/usr/bin/env bash
+# cc_to_srt_safe.sh
+# Convert EIA-608 hex dump to readable text (simplified version)
+bash --version
+INPUT_FILE="$1"
+OUTPUT_FILE="$2"
+OUTPUT_FILE_TMP="${OUTPUT_FILE}.tmp"
+
+if [[ -z "$INPUT_FILE" || -z "$OUTPUT_FILE" ]]; then
+    echo "Usage: $0 <input_file> <output_file>"
+    exit 1
+fi
+
+declare -A char_map=(
+    ["20"]=" "
+    ["A1"]="!"
+    ["A2"]="\""
+    ["23"]="#"
+    ["A4"]="$"
+    ["25"]="%"
+    ["26"]="&"
+    ["A7"]="'"
+    ["A8"]="("
+    ["29"]=")"
+    ["2A"]="á"
+    ["AB"]="+"
+    ["2C"]=","
+    ["AD"]="-"
+    ["AE"]="."
+    ["2F"]="/"
+    ["B0"]="0"
+    ["31"]="1"
+    ["32"]="2"
+    ["B3"]="3"
+    ["34"]="4"
+    ["B5"]="5"
+    ["B6"]="6"
+    ["37"]="7"
+    ["38"]="8"
+    ["B9"]="9"
+    ["BA"]=":"
+    ["3B"]=";"
+    ["BC"]="<"
+    ["3D"]="="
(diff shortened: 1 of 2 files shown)


richardpl:
In theory you could do same thing with ass/subtitle and to limited extent with drawtext filter (limited because it have less text styling features then libass powered filters) - and with lot of work, with some runtime commands dynamic insertion, and this could be nodelay-ish solution as conversion to text subtitles and drawing/overlay back over video is going to add additional delay - but for vrecord that is irrelevant. Anyway I think it should be possible to do all of it in realtime, but not yet with just ffmpeg tool alone. 

When using libraries directly one could insert new draw command to filter on each frame displayed, and code using libraries would  be alone responsible for converting cc output to actual captions/formated text, and this is even more work than your current working "demo" solution here. As you all know subtitle media filtering/filters is still far away from ffmpeg, even though there is some code posted on mailing list months ago, but no consensus where ever reached thus changes where never merged. I never got motivation to use such changes in my fork because of possible API/ABI conflicts in future version of ffmpeg code ...

Its also possible to do new filter which would do CC decoding inside filter and than rendering within that same filter, even more work...

So, I'm afraid there is no any shortcuts here, even thought I would like to have one, but just fully CC text dechipering into full actual .ass styling is not a trivial work for real-time, on the fly CC display. For off-line, non real-time you just decode all frames, filter out relevant lines, collect all CC data generate file and give it to demuxer/decoder combo, but you all know this boring stuff already...

dericed:
I'm realizing that this approach of running `cchex_to_display` before the capture starts and stopping it after also fits in with how we could run a background process to start and stop vrecord conditionally. For example, we could have this script make the capture preview but it could also occasionally check for a timecode value or if the video went to black and stop the capture. I suggest let's finish up this PR first but then I could build off this for https://github.com/amiaopensource/vrecord/issues/499

privatezero:
In theory that would just involve removing it from the list of available options that generate the drop down list for capture view modes, right?

dericed:
@bturkus able to test as well?

iamdamosuzuki:
here's a log from another test. I dropped two frames at the very beginning. I think it may have been caused by me pressing the right arrow key, which seems to have confused the decoder. Not standard operating procedure, so probably not a big deal. 
[20250929_CaptionViewer_01_vrecord_input.log](https://github.com/user-attachments/files/22599269/20250929_CaptionViewer_01_vrecord_input.log)

I do think we should consider just having this view available in passthrough and not capture mode

dericed:
Having it in passthrough only could work, but one opportunity in this approach is that we could use the same intermediate log that gathers caption data to also include timecode and some qctools basis data which could permit us to conditionally start and stop the capture.

PR #936, opened 2025-09-18
compress a huge file

dericed:
This mp4 was a huge proportion of the vrecord download package


retokromer:
@dericed Side note: If you update the documentation, could you possibly also update the version it refers to?

https://github.com/amiaopensource/vrecord/blob/95bfb5b53fbddcacceb09dc8bcb832195007d6aa/README.md?plain=1#L3

privatezero:
Hah! Yeah that didn't need to be that big in the package!

privatezero (approved):

Issue #879, opened 2025-03-28
Add option to include FADGI WebVTT header info for caption data


kmurmur:
Add the option to include the FADGI-defined lightweight metadata tags to give context and define provenance of caption data.  The fields and values help patrons and collection administrators understand what kind of data a WebVTT contains, from where it came, and when, where, and how it was made.

[Guidelines for Embedding Metadata in WebVTT Files](https://www.digitizationguidelines.gov/guidelines/FADGI_WebVTT_embed_guidelines_v1.0.pdf), (PDF 153 KB). Version 1.0. June 7, 2024.

iamdamosuzuki:
I finally got a chance to look at this, it looks great. Here are a few thoughts

- Can you add in `Originating File` and have it fill in with the Preservation file's filename?
- Until the spec revision for `File Creator` being repeatable, there could be a conditional where if `${TECHNICIAN}` is filled out it uses that field, and if `${TECHNICIAN}` is blank it uses `ffmpeg`
- I understand @cmhosale 's point that with English language content the SAP captions would technically be "subtitles". I think we can couch this for now since I don't believe the SAP track is being extracted currently. Is that correct @dericed ?

kmurmur:
So great to see all of this! Thank you! 
- I'm not sure that  File Creator is intended to be Technician since there could be a creating organization. The definition says "If individual: last name [comma space] first name. If organization: free text." 
- I do see an action here to update the File Creator examples to include one for software (aka ffmpeg as an example).
- And yes, I see the point about repeatability for File Creator. So we'll add that as well to an errata list for a potential update.

iamdamosuzuki:
Hi Dave,

I've had a change to look at this. Everything in here looks good, but I would like to request additional fields:

**Responsible Party** - This isn't possible to get from the current info entered by users in vrecord, so there would need to be an additional tab under the Config tab to add this field. My suggestion is that the tab be called WebVTT Metadata.

**Media Identifier** - We already have the filename without extension being used as the Media Identifier, which is fine. However, this would be enhanced if we could add the "type". The additionally WebVTT Metadata tab should have a field called _**Media Identifier Type**_ that is appended after a command and space to the filename in the Media Identifier field. If this field is left blank, the Media Identifier is left as just the file name.

iamdamosuzuki:
Hi @kmurmur. 

[...]

PR #948, opened 2026-01-09
add the fadgi header to webvtt

dericed:
Addresses #879


dericed:
ready for re-review. I put an output header and screenshot on https://github.com/amiaopensource/vrecord/issues/879#issuecomment-3761578863

privatezero (approved):
Can't fully test rn, but seems to work for me in a quick check

Issue #869, opened 2025-02-28
Additional Post-Process Wishlist


iamdamosuzuki:
Here's a list of desired post-processes. I invite anyone add to this list
 
- Wrap timecode in Preservation File (priority for FADGI)
- Wrap caption data in Preservation File
- Create file-level checksum of Preservation File
- Wrap timecode in Access File

iamdamosuzuki:
I would prefer the output of `md5 -q` to avoid the relative path. thanks

PR #949, opened 2026-01-16
Post process subtitle wrap

dericed:
This addresses the last item in https://github.com/amiaopensource/vrecord/issues/869#issuecomment-3725400741 and embeds the captions into the access mp4 as a subtitle track.
Doing so with the preservation file will be a bit more complex since the approach could vary per format and the post checks should be more careful.


retokromer:
Works on macOS 26.2 and 15.7.3 on my end.

retokromer (approved):

privatezero (approved):
Can't fully test, but I _think_ `-E` does the same thing in `sed` on Mac and Linux, so lgtm on paper

privatezero:
@dericed I just tested the above example and it behaves as expected on my end in Ubuntu

PR #951, opened 2026-01-22
mac: Script for creating a .pkg installer

g-maxime:


privatezero:
With recent changes, this seems to perform as expected on my end at least

retokromer (approved):
After installing DeckLink separately it works for me on macOS 26.2 and 15.7.3.

PR #953, opened 2026-02-04
remove option for slice count

dericed:
i'm curious what people think but this PR remove the option to set the ffv1 slice count. Instead it would just default to what ffmpeg picks, which is 4 for SD (720x486) and 20 for HD (1920x1080). Otherwise since we now manage HD and SD, having one global slice count option doesn't make sense.

Alternatively is we don't like the default ffmpeg slice counts, then we could set our own pick for each video standard, like 12 for SD and 30 for HD? Either way I think either setting a vrecord community selected slice count per standard or going with the ffmpeg defaults would be preferable to a user-selected slice count.

Thoughts?


retokromer (approved):
+1 for removing this parameter and using FFmpeg’s default values

bturkus:
I also agree, think this is the cleanest approach!

PR #971, opened 2026-03-03
Access mp4 options

dericed:
this just fine tunes the openh264 encoding a bit and silences a few warnings from the encoder


retokromer (approved):
not tested, but code LGTM

privatezero (approved):
lgtm

PR #975, opened 2026-03-06
use `DEVICE_NAME`

retokromer:


dericed:
I'd rather not. I think this information is useful to log more clearly than it is at the moment, where it exists in the logged record command line. Merging this would set us further back on this, so I'd rather leave it for future use than dig back in here to recover it later.

retokromer:
Why not use it then?

dericed:
@retokromer I suppose it's a matter or timing and scope.

PR #977, opened 2026-03-11
correct a dvpackager label

dericed:
DV_RESCUE_OPTION_TC enables `-S` in dvpackager which is about embedded the extracted caption data as a subtitle. The label here was incorrect. I opted to leave the variable the same so that users that have been using `-S` don't need to reconfigure this.


privatezero (approved):
lgtm!

PR #983, opened 2026-03-31
Added gop size 12 to access mp4

iamdamosuzuki:
added gop size 12 (-g 12) to the access mp4 creation to avoid MP4s with gigantic GOP sizes. I tested this to make sure the FFV1 files were still -g 1 and it appears they are. It'd be nice if someone else could check that out too.


dericed (approved):

privatezero:
This good to be merged?

retokromer (approved):

PR #986, opened 2026-04-03
Update linux docs

privatezero:
Did a quick cleanup of the Linux install docs to remove some outdated advice and to include some other things that have come up.

* Adds `bison` to install list
* Removes note about uninstalling sdl2 as I don't think this is an issue anymore
* Removes instructions about mpv PPA install as I think normal package managers should work fine for this now

### DIFF
--- a/Resources/Documentation/linux_installation.md
+++ b/Resources/Documentation/linux_installation.md
@@ -21,6 +21,7 @@ The following instructions aim to minimize use of Homebrew installs for packages
   - `sudo apt-get install mkvtoolnix`
   - `sudo apt-get install mediaconch`
   - `sudo apt-get install mpv`
+  - `sudo apt-get install bison`
 * Install the following dependencies for enabling DV capture in vrecord's FFmpeg build:
   - `sudo apt-get install libiec61883-dev`
   - `sudo apt-get install libraw1394-dev`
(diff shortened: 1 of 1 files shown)


Resources/Documentation/linux_installation.md
```diff
@@ -27,6 +20,8 @@ The following instructions aim to minimize use of Homebrew installs for packages
   - `sudo apt-get install xmlstarlet`
   - `sudo apt-get install mkvtoolnix`
   - `sudo apt-get install mediaconch`
+  - `sudo apt-get install mpv`
```
retokromer, line 23: Does the official Homebrew formula not work? I am asking, because we use Homebrew on Debian and Ubuntu, but with our own formula.

retokromer (approved):
Only a comment, otherwise it LGTM.

Resources/Documentation/linux_installation.md
privatezero: It's honestly been a while since I tested the official brew formula. Since one of the main headaches for installing vrecord on linux seems to come from brew issues, my preference has tended to be to get as much installed via non-brew package managers as possible, but it is very conceivable that the brew formula would work just fine.

retokromer: Thank you for the answer! Since we have to install some dependencies via `apt-get` anyway, it doesn't really matter.

privatezero:
I'm going to make some further changes to this to help make things a bit more clear for an install on a clean version of Ubuntu

PR #991, opened 2026-05-14
Update Linux install instructions

privatezero:
Updated Linux install docs to simplify the process and remove out of date information - this process is working in my VM test so far.

### DIFF
--- a/Resources/Documentation/linux_installation.md
+++ b/Resources/Documentation/linux_installation.md
@@ -1,67 +1,38 @@
 # Steps for a successful install of Vrecord on Linux
 
 ## About
-There are many possible ways to install the various dependencies of Vrecord on Linux. The main body of this documentation is oriented towards installing on Ubuntu LTS. For Some information about installing on additional distributions see the section at the bottom of this document. 
+There are many possible ways to install the various dependencies of Vrecord on Linux. The main body of this documentation is oriented towards Ubuntu LTS, and has been tested to work on fresh installations of Ubuntu 24.04 LTS and 26.04 LTS.
 
-## via Homebrew
-The following instructions aim to minimize use of Homebrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord.
+### Install non-homebrew core dependencies
+1) If working on a fresh installation of Ubuntu, it is recommended to run `sudo apt-get update && sudo apt-get upgrade -y` prior to installing vrecord to make sure that all packages are their latest versions.
+2) Download and install the latest Linux version of 'Blackmagic Desktop Video' from the [Blackmagic website](https://www.blackmagicdesign.com/support/)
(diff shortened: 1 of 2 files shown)


Resources/Documentation/linux_installation.md
```diff
@@ -1,67 +1,38 @@
...
 
-## via Homebrew
-The following instructions aim to minimize use of Homebrew installs for packages that can otherwise be installed via native Linux methods. When followed in order, these commands should result in a fully functional install of vrecord.
+### Install non-homebrew core dependencies
+1) If working on a fresh installation of Ubuntu, it is recommended to run `sudo apt-get update && sudo apt-get upgrade -y` prior to installing vrecord to make sure that all packages are their latest versions.
+2) Download and install the latest Linux version of 'Blackmagic Desktop Video' from the [Blackmagic website](https://www.blackmagicdesign.com/support/)
+3) Install dependencies required for a successful build. These can be installed one-by-one from the list in [this script](https://github.com/amiaopensource/vrecord/blob/main/Release/linux-dependencies.sh), or that script can be run remotely to simplify installation.
 
-### Programs to be installed manually
+To do this, make sure `curl` is installed (and if not, run `sudo apt install curl`). Then running the following command will install vrecord dependencies:
```
retokromer, line 12: You could also add a test in the `linux-dependencies.sh‎` script and, if not already installed, do it.

retokromer (commented):
one nit, but LGTM

Resources/Documentation/linux_installation.md
privatezero: Bit of a catch-22 though, because ideally people will use `curl` to run linux-dependencies.sh, so can't really have it run a check to see if `curl` is present if they need `curl` to run it.

retokromer: Indeed! Then my nit is void.

retokromer: If `curl` is not yet installed, then `wget` would also do the trick:
```
wget -O - https://raw.githubusercontent.com/amiaopensource/vrecord/refs/heads/linux-doc-updates/Release/linux-dependencies.sh | bash
```
I guess, `wget` comes with Ubuntu? If that is correct, then `wget` could check and when needed install `curl`.

(I personally prefer `curl` for many reasons.)

privatezero: Good point - wget _is_ included out of the box. I think I'll shift the recommended command here to `wget` and then add curl into the dependencies script. I don't think it's even necessary to do a check really, as `apt-get` will just skip it if it is already installed. Either way, `curl` is necessary for running the following Homebrew install step, so it will have to be installed one way or another.

privatezero: Just made that change and squished it - thanks for the suggestion!

retokromer: Good point indeed, `apt-get` needs it!

retokromer (approved):

privatezero:
@dericed @bturkus, anyone else have thoughts about this? Good to merge?

dericed (approved):
LGTM. thx for all this documentation

Issue #995, opened 2026-06-03
MediaInfo CodecID and MaxSlicesCount mismatch on recently captured FFV1 files


BleakleyMcD:
QC against MediaInfo is failing on freshly captured files. The CodecID is being reported in the Matroska-native form rather than the expected VFW/FOURCC form, and the FFV1 slice count is 4 instead of the expected 24.
```
MediaInfo output from QC:
Some specified MediaInfo fields or values are missing or don't match:
Metadata field CodecID has a value of: V_FFV1
The expected value is: V_MS/VFW/FOURCC / FFV1
Metadata field MaxSlicesCount has a value of: 4
The expected value is: 24
```

JPC_AV_03796_capture_options.log

```
computer_name: Mac-68
computer_model_name: Mac mini
computer_model_id: Mac16,10
computer_processor_name: 
computer_processor_speed: 
computer_processor_count: 
computer_memory: 16 GB
computer_serial: RL4K4FXKGK
computer_cores: 10
user_name: digitizer_a
operating_system_VERSION: Darwin Kernel Version 24.3.0: Thu Jan  2 20:22:58 PST 2025; root:xnu-11215.81.4~3/RELEASE_ARM64_T8132
vrecord version: 2026-03-02
datetime_start: 2026-06-01T13:36:43
capture_device_node: /dev/disk3
capture_device_protocol: Apple Fabric
capture_device_name: APPLE SSD AP0256Z
capture_device_block_size: 4096 Bytes
capture_device_location: Internal
capture_device_partition_type: 41504653-0000-11AA-AA11-00306543ECAC
capture_device_partition_name: APFS
capture_device_volume_uuid: DC2C1AD8-D4FF-4D36-9376-524CB49BE049
capture_device_volume_available: 
capture_device_volume_total: 
FILE_PATH: /Users/digitizer_a/Desktop/Ingestion/JPC_AV_03796/JPC_AV_03796.mkv
video_card_name: UltraStudio 4K Mini
VIDEO_INPUT_CHOICE: Composite
AUDIO_INPUT_CHOICE: Analog
VIDEO_BIT_DEPTH_CHOICE: 10 bit
AUDIO_MAPPING_CHOICE: 2 Stereo Tracks (Channels 1 & 2 -> 1st Track Stereo, Channels 3 & 4 -> 2nd Track Stereo)
STANDARD_CHOICE: NTSC
ASPECT_RATIO_CHOICE: 4/3
[...]

PR #996, opened 2026-06-09
setting recommended slice count per frame format

dericed:
responds to discussions at https://github.com/amiaopensource/vrecord/issues/995


privatezero (approved):
lgtm!

retokromer (approved):
For ordinary computers, I would personally have chosen smaller numbers, but I like this approach.

PR #997, opened 2026-06-23
don't use timecode_format on faux inputs

dericed:
This fixes an issue when running passthrough on a file input.


privatezero (approved):
works great - thanks!

retokromer (approved):

PR #1007, opened 2026-07-22
Add "analog rca" and "analog xlr" audio inputs

stith:
On my older UltraStudio 4 (thunderbolt 2 edition), these inputs are chosen separately.

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -1928,6 +1928,8 @@ _lookup_choice(){
 
         # audio inputs
         "Analog")                  AUDIO_INPUT="analog" ;;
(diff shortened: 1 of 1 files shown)


dericed:
I think this particular arrangement is pretty rare. Maybe we need something like `if device='UltraStudio 4' ; then display the extra options`. @privatezero?

retokromer:
>I think this particular arrangement is pretty rare.

I agree.

privatezero:
@dericed - I agree it probably would be ideal to not display the extra options for devices that they aren't compatible with (which is the majority of devices).

PR #1010, opened 2026-08-25
Mediaconch policy updates and adding mp4 tests

dericed:


iamdamosuzuki:
As mentioned in my review of #1011, files are failing colour_primaries and matrix_coefficients because those fields are blank in the file. I think the best bet is to remove those rules for the access files, unless there's a way to force ffmpeg to write those fields properly for the MP4

PR #1011, opened 2026-08-25
Fix access mp4 framerate

dericed:


dericed:
this PR is based off https://github.com/amiaopensource/vrecord/pull/1010

iamdamosuzuki (approved):
Framerate now comes out at 29.97, so looks good. Still failing media conch because of color primaries and matrix coefficients because those fields are empty in the MP4. Maybe it's best to delete those rules in #1010 ?

privatezero (approved):
this seems to work on my end (tested with a file input, not live capture). I didn't have a problem with the matrix coefficients for the MP4 weirdly. I did have those failures when my output was set to AVI though (values are blank in the file)

PR #1015, opened 2026-08-28
add mediaconch tests for aspect ratio

### LINKED ISSUE #970: add aspect ratio tests to the mediaconch policy post-check

dericed:
resolves https://github.com/amiaopensource/vrecord/issues/970
This was a little tricky since mediaconch doesn't give a way to test numerator and denominator values without specifying the trace location for each container (cc @jeromemartinez). So here I decided to convert the fractional SAR (or DAR) to a decimal and test if the stored value is within a small range around it (to forgive rounding errors).


privatezero (approved):
only tested with virtual input, but seems OK on my end!

PR #1016, opened 2026-08-28
skip the unprocessed eia608 output

### LINKED ISSUE #962: no need to keep the temp eia608 file around
Currently it's saved if captions are found, but once the scc file is made it's not needed.

dericed:
resolves https://github.com/amiaopensource/vrecord/issues/962


privatezero (approved):
lgtm

PR #1018, opened 2026-08-29
need to clarify the signal input's pixel format

dericed:
Hi @privatezero, this removes a few warnings. Since the refactor to a single tee muxer output arguments need to include stream specifiers to keep things straight.


privatezero (approved):
looks good to me!

retokromer (approved):

PR #1019, opened 2026-09-01
Color free terminal alerts

privatezero:
This adds an option in settings tab to disable usage of color in vrecords terminal alerts. Warnings are mapped as follows:
* red -> inverted background/bold
* blue -> inverted background/italics
* green -> inverted background

### DIFF
--- a/vrecord
+++ b/vrecord
@@ -74,11 +74,19 @@ _get_iso8601(){
 }
 
 _report(){
-    local RED="$(tput setaf 1)"    # Red      - For Warnings
-    local BG_RED="$(tput setab 1)"    # Background Red      - For STRONG Warnings
-    local GREEN="$(tput setaf 2)"  # Green    - For Declarations
-    local BLUE="$(tput setaf 4)"   # Blue     - For Questions
-    local NC="$(tput sgr0)"        # No Color
+    if [[ "${COLOR_SAFE}" == 'true' ]] ; then
+        local RED="$(tput bold)$(tput smso)"   # Red      - For Warnings
+        local BG_RED="$(tput setab 1)"    # Background Red      - For STRONG Warnings
+        local GREEN="$(tput smso)"  # Green    - For Declarations
+        local BLUE=""$(tput sitm)$(tput smso)""   # Blue     - For Questions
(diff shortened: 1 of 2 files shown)


vrecord
```diff
@@ -74,11 +74,19 @@ _get_iso8601(){
...
+        local GREEN="$(tput smso)"  # Green    - For Declarations
+        local BLUE=""$(tput sitm)$(tput smso)""   # Blue     - For Questions
+        local NC="$(tput rmso)$(tput ritm)"        # No Color
+    else
+        local RED="$(tput setaf 1)"    # Red      - For Warnings
+        local BG_RED="$(tput setab 1)"    # Background Red      - For STRONG Warnings
+        local GREEN="$(tput setaf 2)"  # Green    - For Declarations
+        local BLUE="$(tput setaf 4)"   # Blue     - For Questions
+        local NC="$(tput sgr0)"        # No Color
+    fi
```
retokromer, line 90: Two nits:
- I would align at # for readability
- Use `inverted background` etc. in the comment when the colours are replaced.

privatezero: good call! I didn't even look at the comments :-)

privatezero:
Just in case anyone is wondering, I left the Background Red option in there as-is since it isn't currently used in vrecord. Should it become used again we can update it, but figured it was better to have it in there so it wouldn't just throw an error in case nobody remembered to change it.

retokromer (approved):

dericed:
Hi @privatezero I added a commit to update the $NC (no color) variable. The one in non-color-safe worked but the one you used in the color safe option would leave the terminal impacted after vrecord ended (like the terminal would stay in bold). I also added a commit to drop color from ffmpeg/ffplay as well. Feel welcome to drop or keep.

dericed (approved):

privatezero:
> Hi @privatezero I added a commit to update the $NC (no color) variable. The one in non-color-safe worked but the one you used in the color safe option would leave the terminal impacted after vrecord ended (like the terminal would stay in bold). 

Oh interesting! Must be another system difference there, both work on my end. Thanks for adding those!

Issue #989, opened 2026-05-08
More options for file input mode and/or qc mode


privatezero:
I use vrecord file input mode as part of my QC process for files received from batch digitization vendors. Making an issue here as a space to remind myself to investigate adding some added features/functionality here around using vrecord for QC. Some possible thoughts include:
* adding some options to better support integrating vrecord into QC loop scripts, such as:
  * Allow specified length to play input file for (e.g. 60 seconds) -  potentially could leverage existing variable for setting transfer length
  * Allow randomized start time for chunk of file to play from file input (so not all files start playing at beginning)
* Use `drawtext` to display file name and some technical metadata (bit depth, codec etc)

iamdamosuzuki:
Hi @privatezero, these are great ideas. I would like to suggest that there be a playback mode that plays the first 15 seconds, a random 15 second chunk in the middle, and the last 15 seconds. Or maybe there be some methods by which it would play an arbitrary number of 15 second chunks in the middle.

iamdamosuzuki:
Here are some more wishlist features for a QC mode

- the ability to turn on or off different filters, or maybe the ability to choose up to four? Ideally I'd like to see scopes, broadcast highligo, and bitplane slice. I'm sure other folks have preferences

- I'm not sure this is possible with bash, but it'd be cool to bind a key to "fail" a file, which could tag the file with the red color on the macos finder color tagging system. So if you run a folder full of files through qc mode you can quickly see which files you failed after the fact

privatezero:
I've started to do some work on this [here](https://github.com/amiaopensource/vrecord/tree/update-lua) as a first step. I've changed the existing lua script so that it uses `--lavfi-complex` for all filters rather than `lavfi` which allows the combination of video and audio scopes. A key difference from ffplay is this allows scrubbing! I'll tweak it all from here and see about integrating some of the other options/filters.

Issue #847, opened 2024-10-28
Non-Responsive Key Bindings for Visual + Numerical Filters (mpv QC View, #3 and #4)


bturkus:
**Summary**

Key bindings 3 and 4 for the "Visual + Numerical" views in the mpv Lua script are currently unresponsive. When pressing these keys, no visualizations appear, despite other key bindings working as expected. Both views combine visual elements with associated numerical scores, intended to display broadcast and full-range views, respectively.

**Details**

Affected Key Bindings: 3 ("Visual + Numerical: Broadcast Range") and 4 ("Visual + Numerical: Full Range").

Expected Behavior: Display a composite of visual waveform and vectorscope views, with numerical range indicators for each.

Current Behavior: No response or visualization on pressing keys 3 and 4.

**Related Issue**

This issue is somewhat similar to an existing [Audio Passthrough issue (#782)](https://github.com/amiaopensource/vrecord/issues/782), where audio passthrough functionality remains inactive due to a lavfi-complex requirement.

Thanks!

Ben

Issue #782, opened 2024-01-31
Audio Passthrough Keybinding toggle doesn't work


marshalleq:
Other key bindings do, including volume adjustment, but this one does not.  Anyone else have it working?  Supposedly just pressing a will toggle it right?

privatezero:
@marshalleq the audio passthrough mode started by running `vrecord -a` is basically just a quick way to get vrecord to fire up into the "Audio + Video" preview window - could you confirm what mode you are running and how it isn't working?

marshalleq:
I'm just running vrecord with no config options (I guess some were saved previously).  After I've started recording there are a number of keybindings I can use, one of which is the a key, shown on the on screen display.  My expectation is that I can press that to mute the audio being passed through from the source.  This would be useful as I actually have meetings at the same time on the same machine and it can be distracting not being able to mute it.  Probably not that many archivists do this quite like this, but machines are so powerful now.  Of course there are other reasons one would want to mute it also.

privatezero:
Thanks for clarifying! The vrecord preview window uses ffplay, so it is limited to the [commands available](https://ffmpeg.org/ffplay.html#While-playing) to ffplay. Using `m` during passthrough seems to be working for muting on my system - does this not work while you are capturing?

marshalleq:
Hmmm, interesting, I was going from the below, (see OSD screenshot), which does not have an m option, though I've now confirmed that does work.  So what is the 'a' option for?  Is this just out of date, or does it stop audio going to the recording?  A few others seem to do nothing also.

Thanks.
<img width="613" alt="Screenshot 2024-02-01 at 13 03 20" src="https://github.com/amiaopensource/vrecord/assets/13308996/22b6b3ff-2bea-46fc-a6fb-417a3873bef6">

privatezero:
Ah - you are running in 'Quality Control' view - that's actually a bit different from how the other views work. QC View uses mpv rather than ffplay for the window which is what gives it a bit more interactivity. But yes, it was a bit of a work in progress I think so some of the options might not actually be functional which is why `a` isn't doing anything. 

The audio mode that can be started either with `vrecord -a` or by selecting it in the settings (Audio + Video) basically is just a view that adds in some audio scopes like in [this image](https://github.com/amiaopensource/vrecord/blob/main/Resources/audioPlusVideoMode-small.png). It used to be referred to as 'Audio Passthrough' as it was only available in passthrough mode, so that's probably why the broken toggle in QC View is named like that.

PR #1021, opened 2026-09-03
Update lua script for mpv based qc view

privatezero:
Switches the qc view lua script to use --lafi-complex for filters and fixes issues with audio scopes and signal stats.

This addresses https://github.com/amiaopensource/vrecord/issues/847 and https://github.com/amiaopensource/vrecord/issues/782 and can be built upon for ideas in https://github.com/amiaopensource/vrecord/issues/989.

Note that for Video + Signal view I dropped the stats being displayed at the bottom of the screen to make things cleaner (and I wasn't sure how to access the main vrecord log path variable from within the lua script since it is assigned externally).


dericed:
@privatezero have you ran this with a live source. When I play from tape I get a half second pause in the player every 4 seconds or so

privatezero:
@dericed I'm having some issues on my mac mpv build with it not linking to dependencies correctly to enable `drawtext` so I haven't been able to test out all the views, but in the ones that don't involve `drawtext` I don't have any pauses with a live source.

Is there a particular view that is triggering that for you, or is it all of them?

retokromer:
Hello @privatezero 
Out of curiosity: what `mpv` are you using? I am asking, because since a while now we use our own flavour, which is linked to our `ffmpeg` implementation. If I recall carefully, currently `vrecord` uses `ffmpeg-ma`.

privatezero:
@retokromer it's the one that gets built by brew - there are a few things tangled up on that computer, so I just assume that it isn't linking to `ffmpeg-ma` for whatever reason. I'll just need to go sort out what is happening over there - I have no similar problems on my Ubuntu machine using standard builds.

retokromer:
Thank you! (If an AMIA implementation is needed, don't hesitate to get in touch.)

privatezero:
@dericed for expediency of testing I loaded a different build of MPV (mac ports installation) and retested - I still didn't have any pauses or issues during both live preview and live capture, even while using Audio + Video mode, which should be the most intensive filter graph in there

